PR #3, opened 2018-11-12
chore(gitignore): Ignore swap files

jrmykolyn:


.gitignore
```diff
@@ -13,6 +13,7 @@ packages/*/test/**/*.d.ts
 
 package-lock.json
 *.log
+**/.*.swp
```
jmbtutor: A comprehensive ignore pattern would actually be this:
```
[._]*.s[a-w][a-z]                                                                                                                                                                                                                                                              
[._]s[a-w][a-z]
```

jrmykolyn: Do you have some examples of other swap-type files which these patterns would match?

Also: RIP PR, we hardly knew ye.

jmbtutor: If vim sees that a swap file exists and you tell it to edit anyway, it will create a `.swo` file, and it goes up the alphabet from `p` if even that is present.

PR #5, opened 2018-11-13
SF-1132: Default to an empty array for refinements, and pass through the rest of parameters from a request

Vic-Dev:


packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,9 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [<release type>]
```
jmbtutor: Use the proper release type. This is probably a `patch`.

packages/@storefront/core/src/services/urlUtils.ts
```diff
@@ -136,15 +137,15 @@ namespace UrlUtils {
   // tslint:disable-next-line max-line-length
   export const getAllIds = (state: Store.Indexed<Store.Navigation> | Store.AvailableNavigations, request: UrlBeautifier.SearchUrlState) => {
     return state.allIds.concat(
-      ...request.refinements.map(({ field }) => field).filter((field) => !state.allIds.includes(field))
+      ...(request.refinements || []).map(({ field }) => field).filter((field) => !state.allIds.includes(field))
```
jmbtutor: We don't use anything else from `request`, so `refinements` can be destructured and given a default in the function declaration.

packages/@storefront/core/src/services/urlUtils.ts
```diff
@@ -136,15 +137,15 @@ namespace UrlUtils {
...
     );
   };
 
   // tslint:disable-next-line max-line-length
   export const getById = (state: Store.Indexed<Store.Navigation> | Store.AvailableNavigations, request: UrlBeautifier.SearchUrlState) => {
     const byId = { ...state.byId };
 
-    request.refinements.forEach((refinement) => {
+    (request.refinements || []).forEach((refinement) => {
```
jmbtutor: Same thing here.

packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,9 +4,14 @@ All notable changes to this project will be documented in this file.
...
+
```
jmbtutor: I don't think we have a newline here in our format, but it doesn't really matter.

PR #18, opened 2018-11-29
SF-1196: Use Navigation Name for Toggle Refinement Crumbs

JoeyDeol:


packages/@storefront/breadcrumbs/test/unit/refinement-crumb.ts
```diff
@@ -0,0 +1,95 @@
...
+    it('should set state', () => {
+      const label = 'test';
+      const props = refinementCrumb.props = <any>{ a: 'b', c: 'd', range: false, value: label };
+
+      refinementCrumb.updateState();
+
+      expect(refinementCrumb.state).to.eql({ ...props, label });
+    });
+
+    it('should set label to a range', () => {
```
jrmykolyn: What happens if range-type refinement doesn't include both a `low` and `high` value?

packages/@storefront/core/src/core/defaults.ts
```diff
@@ -91,6 +91,9 @@ const DEFAULTS: Partial<Configuration> = {
   refinements: {
     overrides: {},
   },
+  navigations: {
+    type: {},
```
jrmykolyn: Is the `type` key reserved for 'toggle'/'boolean'-type navigations? If so, is there a more semantic key?

packages/@storefront/flux-capacitor/test/unit/core/adapters/configuration.ts
```diff
@@ -280,6 +280,24 @@ suite('Configuration Adapter', ({ expect, stub }) => {
     });
   });
 
+  describe('extractToggleNavigations()', () => {
+    it('should return the toggle navigations', () => {
```
jrmykolyn: Is there a 'type' of navigations that this function should ignore? If so, lets test for that.

packages/@storefront/flux-capacitor/test/unit/core/adapters/search.ts
```diff
@@ -230,35 +231,63 @@ suite('Search Adapter', ({ expect, stub }) => {
...
+        { field: 'Limits to max', refinements: [0, 1, 2, 3, 4], selected: [], show: [0, 1], more: true, boolean: false },
       ]);
+      // tslint:enable
       expect(max).to.be.calledWithExactly(conf);
       expect(config).to.be.calledWithExactly(state);
     });
 
+    it('should mark which navigations are boolean type', () => {
+      const navigations: any = [
+        { field: 'one', refinements: [2, 4], selected: [], more: false },
```
jrmykolyn: Consider pulling these navigation objects into variables which can be spread and enriched in the assertion below. ie:

```
const nav1 = { ... };
const nav2 = { ... };

const navigations = [nav1, nav2, ...]

...

 expect(Adapter.pruneRefinements(navigations, state)).to.eql([
{ ...nav1, boolean: false },
{ ...nav2, boolean: false },
...
]);
```

packages/@storefront/core/src/core/defaults.ts
JoeyDeol: type is set up to be future-proofed so we can add additional types of navigations beyond just toggle/boolean-type.

packages/@storefront/flux-capacitor/src/core/adapters/configuration.ts
```diff
@@ -163,6 +163,9 @@ namespace Adapter {
   export const extractRefinementsSort = (config: Configuration) =>
     extractINav(config).refinements.sort;
 
+  export const extractToggleNavigations = (config: Configuration) =>
+    Object.keys(config.navigations.type);
```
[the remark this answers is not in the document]
JoeyDeol: Updated to only pull in navigations that have a value of 'toggle'

packages/@storefront/breadcrumbs/test/unit/refinement-crumb.ts
jmbtutor: We don't do that assertion anywhere else, so I would say it's out of scope. If we were to do it, we'd probably want to do it for everything else as well.

packages/@storefront/flux-capacitor/src/core/adapters/search.ts
```diff
@@ -86,21 +87,32 @@ namespace Adapter {
...
+        for (let i = 0; i < navigation.refinements.length && show.length < max; i++) {
+          if (!navigation.selected.includes(i)) {
+            show.push(i);
+          }
         }
-      }
 
-      return {
-        ...navigation,
+        return {
```
jmbtutor: Since `{ ...navigation, boolean: bool }` is the same between both cases, perhaps it's worth it to extract that into a variable and add `more` and `show` only if `max` exists, then returning the resulting navigation. This would also eliminate the `else` case and make it more flexible in case we need to add more properties in the future.

packages/@storefront/flux-capacitor/test/unit/core/adapters/configuration.ts
```diff
@@ -5,6 +5,7 @@ import { DEFAULT_COLLECTION } from '../../../../src/core/reducers/data/collectio
 import * as PageReducer from '../../../../src/core/reducers/data/page';
 import * as PastPurchaseReducer from '../../../../src/core/reducers/data/past-purchases';
 import suite from '../../_suite';
+import { ADDRCONFIG } from 'dns';
```
jmbtutor: Remove.

packages/@storefront/flux-capacitor/test/unit/core/adapters/search.ts
```diff
@@ -230,35 +231,63 @@ suite('Search Adapter', ({ expect, stub }) => {
...
-        { name: 'Limits to max', refinements: [0, 1, 2, 3, 4], selected: [], more: true },
+        { field: 'Adds non-selected refinement', refinements: [4], selected: [], more: false },
+        { field: 'Adds selected refinements', refinements: [6, 7, 4, 5, 6, 8], selected: [1, 5] },
+        { field: 'Adds selected refinement first', refinements: [8, 9], selected: [1], more: true },
+        { field: 'Limits to max', refinements: [0, 1, 2, 3, 4], selected: [1, 2, 3, 4], more: true },
+        { field: 'Limits to max', refinements: [0, 1, 2, 3, 4], selected: [], more: true },
       ];
+      stub(ConfigAdapter, 'extractToggleNavigations').returns([]);
 
+      // tslint:disable
```
jmbtutor: Only disable the rule that you need (probably `max-line-length` in this case).

packages/@storefront/breadcrumbs/src/refinement-crumb/index.ts
```diff
@@ -1,3 +1,49 @@
...
+    if (this.props.range) {
+      label = this.props.low + ' - ' + this.props.high;
+    } else if (this.props.boolean) {
+      label = this.props.navigationLabel;
+    } else {
+      label = this.props.value;
+    }
+
+    this.state = {
+      ...this.props,
```
Vic-Dev: Why are you spreading props into state?

packages/@storefront/flux-capacitor/test/unit/core/adapters/search.ts
Vic-Dev: I wrote it this way so it would disable for multiple lines, to do what you want we need to disable max-line-length on each line

packages/@storefront/breadcrumbs/test/unit/refinement-crumb.ts
Vic-Dev: If a range refinement comes back from the DB, it will always have a `low` and `high`. So a range refinement in the store will always have a `low` and `high`. The only way you could end up in a scenario where it doesn't have those properties is if either someone misuses this component, or someone adds a range refinement to the store manually and doesn't provide `low` & `high`. I don't think we should support these cases.

Vic-Dev: But there's a ticket somewhere about validating props which I think accomplishes what you want in this scenario (+ I'm in the process of moving our normalization to the reducers which means that everything that is created in the store will always have the correct shape and types)

packages/@storefront/flux-capacitor/test/unit/core/adapters/configuration.ts
JoeyDeol: Done.

packages/@storefront/breadcrumbs/src/refinement-crumb/index.ts
JoeyDeol: Reverted this changes and made it so that the component doesn't require state anymore to do the logic checks.

packages/@storefront/flux-capacitor/test/unit/core/adapters/search.ts
JoeyDeol: `// tslint:disable max-line-length` seems to work to disable multiple lines.

packages/@storefront/breadcrumbs/test/unit/refinement-crumb.ts
jrmykolyn: > But there's a ticket somewhere about validating props which I think accomplishes what you want in this scenario (+ I'm in the process of moving our normalization to the reducers which means that everything that is created in the store will always have the correct shape and types)

Sounds good; marking this as resolved.

packages/@storefront/flux-capacitor/test/unit/core/adapters/configuration.ts
jrmykolyn: Since the update test case includes navigation types *other* than 'toggle', this is addressed.

packages/@storefront/flux-capacitor/test/unit/core/adapters/search.ts
jrmykolyn: Addressed, resolving.

packages/@storefront/breadcrumbs/src/refinement-crumbs/index.ts
```diff
@@ -56,6 +56,8 @@ class RefinementCrumbs {
             index,
             field,
             range,
+            navigationLabel: navigation.label,
```
Vic-Dev: You can still also do your logic for label and add it to the refinements state here. https://github.com/groupby/storefront/blob/4fa4f288c84b1c5a95dff742e3c05dc955127cba/packages/%40storefront/breadcrumbs/src/refinement-crumb/index.ts#L17-L23

Then you can use `props.label` in the `refinementCrumbs` template

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Added
+- SF-1196: Added support for `toggle` type navigations.
+  - The result is that `toggle` type navigations generate refinement crumbs that display the navigation name.
```
jmbtutor: You should mention that a setting in the config was added.

PR #19, opened 2018-12-03
SF-1142: Add CDN release jobs to workflow

jmbtutor:
The supporting JavaScript files were copied almost verbatim from storefront-presets.


webpack.config.js
```diff
@@ -0,0 +1,66 @@
...
+    }),
+  ].concat(ENV === PROD ? [new UglifyJSPlugin({ sourceMap: true })] : []),
+
+  module: {
+    rules: [{
+      test: path.resolve(__dirname, 'presets', 'core.js'),
+      use: { loader: 'expose-loader', options: 'storefront' }
+    }, {
+      resource: [
+        // /node_modules\/@storefront\/.*\.js$/,
```
jrmykolyn: Is this required?

presets/package-versions.js
```diff
@@ -0,0 +1,24 @@
+module.exports = [
```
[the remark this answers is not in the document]
jmbtutor: This file is imported as part of `core.js`, which is included in the resulting bundle. Will reading directly from the filesystem be resolved before it makes it into the bundle?
https://github.com/groupby/storefront/pull/19/files#diff-2f77dc81edaf077954bc0165fa8b4221R5

webpack.config.js
jmbtutor: This was in the file in storefront-presets, but it appears to be left over from an early iteration. I think we can safely remove it.

presets/package-versions.js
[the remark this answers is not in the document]
jrmykolyn: Noted, lets proceed with the current solution.

webpack.config.js
jrmykolyn: Sounds good.

PR #20, opened 2018-12-05
SF-1211: Update `saveState` to accept optional URL

jrmykolyn:


packages/@storefront/core/test/unit/services/url.ts
```diff
@@ -347,6 +364,29 @@ suite('URL Service', ({ expect, spy, stub, itShouldBeCore, itShouldExtendBaseSer
...
+    it('should accept a url override', () => {
+      const currUrl  = '/foo';
+      const route = 'search';
+      const state = { a: 'b' };
+      const urlOverride = '/bar';
+      service['app'] = <any>{ flux: { store: { getState: () => state } } };
+      service['emitUrlUpdated'] = spy();
+      service['filterState'] = spy(() => state);
+      service['handleUrl'] = spy();
+      service['opts'] = { redirects: {} };
```
jmbtutor: Are these using `[]` because of TypeScript? If not, then just use `.`. Additionally, if you don't need to assert that these spies are called, then just replace them with `() => null`.

packages/@storefront/core/test/unit/services/url.ts
```diff
@@ -347,6 +364,29 @@ suite('URL Service', ({ expect, spy, stub, itShouldBeCore, itShouldExtendBaseSer
...
+      win.history = { pushState: spy(() => win.location = { href: urlOverride } ) };
```
jmbtutor: Pull `pushState` into a variable and use that variable in the assertion.

packages/@storefront/core/test/unit/services/url.ts
```diff
@@ -347,6 +364,29 @@ suite('URL Service', ({ expect, spy, stub, itShouldBeCore, itShouldExtendBaseSer
...
+      win.location = { href: currUrl };
+
+      service.updateHistory(<any>{ state: <any>{}, route, url: urlOverride });
+
+      expect(win.history.pushState).to.be.calledWithExactly(
+        { url: urlOverride, state, app: STOREFRONT_APP_ID },
+        '',
+        urlOverride
+      );
+      expect(service['emitUrlUpdated']).to.be.calledWithExactly(currUrl, urlOverride, urlOverride);
```
jmbtutor: Extract the `emitUrlUpdated` spy into a variable.

packages/@storefront/core/src/services/url.ts
```diff
@@ -124,7 +127,8 @@ class UrlService extends BaseService<UrlService.Options> {
   }
 
   buildUrlAndReplaceHistory = ({ state, route }: { state: Store.State, route: string }) => {
-    const url = this.beautifier.build(route, this.urlState[route](state));
+    const url = this.url || this.beautifier.build(route, this.urlState[route](state));
+    delete this.url;
```
jmbtutor: Do we want to be deleting this property, or just setting it to `null` or `undefined`? I would probably do the latter.

jrmykolyn: @jmbtutor Are we concerned that `delete` may inadvertently remove a property on the prototype? If not, is there a reason to assign `undefined`?

jmbtutor: It's not that it would remove the property on the prototype (I don't think that's even possible without explicitly referring to the prototype), but as a matter of semantics, it's that we expect the `url` property to exist but to have some sentinel value. Because of that, I think we should explicitly set it to some sentinel value like `null` or `undefined` (or even `''` if that makes sense) instead of deleting it.

jrmykolyn: @jmbtutor With that in mind, my gut says that setting it `undefined` is closer to the existing solution. That said, to what degree to we need to protect against `undefined` being redefined?

packages/@storefront/core/test/unit/services/url.ts
jrmykolyn: Noted, will update.

packages/@storefront/core/test/unit/services/url.ts
jrmykolyn: Noted, will update.

packages/@storefront/core/test/unit/services/url.ts
jrmykolyn: Noted, will update.

packages/@storefront/core/test/unit/services/url.ts
jrmykolyn: Majority of dynamic property accessors replaced with `.` notation. `['app']` and `['opts']` left for consistency with other tests.

packages/@storefront/core/test/unit/services/url.ts
```diff
@@ -347,6 +364,31 @@ suite('URL Service', ({ expect, spy, stub, itShouldBeCore, itShouldExtendBaseSer
...
     });
 
+    it('should accept a url override', () => {
+      const currUrl  = '/foo';
+      const emitUrlUpdated = spy();
+      const pushState = spy(() => win.location = { href: urlOverride });
+      const route = 'search';
+      const state = { a: 'b' };
+      const urlOverride = '/bar';
+      service.emitUrlUpdated = emitUrlUpdated;
```
jmbtutor: This can be absorbed as a multiple assignment with the initial declaration:
```js
const emitUrlUpdated = service.emitUrlUpdated = spy();
```

jrmykolyn:
This change is no longer required due to https://groupby.atlassian.net/browse/SF-1185. Closing.

PR #28, opened 2018-12-18
SF-1185: Make a URL store

Vic-Dev:


packages/@storefront/flux-capacitor/src/flux-capacitor.ts
```diff
@@ -52,12 +73,74 @@ class FluxCapacitor extends Emitter {
...
+  initHistory({ build, parse, initialUrl, pushState, replaceState }: History) {
+    this.history = {
+      build,
+      initialUrl,
+      pushState,
+      replaceState,
+      noop: () => null,
+      parse: (u) => Promise
+        .resolve(parse(u))
+        .catch((e) => console.warn('UrlService parse promise failed', e)),
```
Vic-Dev: Need to replace this w/ `app.log` somehow

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
```diff
@@ -52,12 +73,74 @@ class FluxCapacitor extends Emitter {
...
+            route,
+            request,
+            url,
+            shouldFetch: urlState.shouldFetch,
+            method: urlState.method
+          })
+        )
+      );
+    } catch (e) {
+      console.warn('unable to parse state from url', e);
```
Vic-Dev: Need to replace this w/ `app.log` somehow

packages/@storefront/core/src/services/url.ts
```diff
@@ -24,166 +20,98 @@ class UrlService extends BaseService<UrlService.Options> {
...
+    const routes = this.opts.routes;
+
+    return Object.keys(routes).reduce(
+      (generatedRoutes: any, prop) => Object.assign(generatedRoutes, { [prop]: basePath + routes[prop] }),
+      {}
+    );
   }
 
-  rewind = (event: PopStateEvent) => {
+  rewind(event: any) {
```
jmbtutor: Can we make this some sort of Event type?

packages/@storefront/core/test/unit/services/url.ts
```diff
@@ -45,541 +63,266 @@ suite('URL Service', ({ expect, spy, stub, itShouldBeCore, itShouldExtendBaseSer
...
   });
 
   describe('init()', () => {
-    it('should listenForHistoryChange and handleUrl', () => {
-      const listenForHistoryChange = service.listenForHistoryChange = spy();
-      const handleUrl = service.handleUrl = spy();
+    it('should call initHistory in flux', () => {
+      const initHistory = spy();
+      const initialUrl = 'www.example.com/ayylmao';
+      const pushState = spy(function() { expect(this).to.eq(service) });
```
jmbtutor: Sinon spies have `thisValue` or something like that that you can assert.

packages/@storefront/core/test/unit/services/url.ts
```diff
@@ -45,541 +63,266 @@ suite('URL Service', ({ expect, spy, stub, itShouldBeCore, itShouldExtendBaseSer
...
-      service.beautifier = <any>{ parse };
-      service['app'] = <any>{ log: { warn } };
-
-      service.handleUrl();
-
-      expect(parse).to.be.calledWith(href);
-      expect(warn).to.be.called;
+      const initHistoryArgs = initHistory.getCall(0).args[0];
+
+      expect(initHistory).to.be.called;
```
jmbtutor: You should expect it to be called first before extracting the args of the first call.

packages/@storefront/flux-capacitor/src/core/actions/payloads.ts
```diff
@@ -2,17 +2,34 @@ import Configuration from '../configuration';
 import Store from '../store';
 
 namespace Payload {
+  export namespace History {
+    export interface State extends PreState {
+      method: Function;
+      request?: any;
+    }
+
+    export interface PreState {
```
jmbtutor: `ProtoState`? What is this actually?

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
```diff
@@ -20,6 +20,23 @@ declare module 'redux' {
   }
 }
 
+// NOTE: `any` temporarily added to suppress compiler warnings re: `void`-type.
```
jmbtutor: Is `void` not a valid type? In any case, make sure there's a ticket for this.

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
```diff
@@ -20,6 +20,23 @@ declare module 'redux' {
...
+// TODO: Remove `any` from union type definition.
+export type Parse = <T>(url: string) => Promise<any | { route: string, request: T }>;
+export type Build = <T>(type: string, state: Store.State) => string;
+
+// TEMP: This was originally used/expoerted by the storefront-core's UrlService.
```
jmbtutor: Mark this as deprecated. You might be able to use the tsdoc `@deprecated` directive.

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
jmbtutor: I think we just use `console` everywhere in flux. The larger ticket would be to move the logger into flux.

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
```diff
@@ -69,6 +70,20 @@ export const UNDOABLE_ACTIONS = [
 export namespace Middleware {
   export const validator = validatorMiddleware();
 
+  export function updateHistory(flux: FluxCapacitor): ReduxMiddleware {
+    return (store: Store<any>) => (next) => (action) => {
+      if (action.type === Actions.UPDATE_HISTORY) {
+        const state = store.getState();
+        const { payload: { method, url = Selectors.url(state) } } = action;
+        const filteredState = filterState(state, action.payload);
```
Vic-Dev: Instead of passing the payload in and using it to build up `filterState` so that it gets the right history, should call the `method` after this action has gone through (ie, need to send the method as an action)

packages/@storefront/core/src/core/defaults.ts
```diff
@@ -99,6 +100,10 @@ const DEFAULTS: Partial<Configuration> = {
   },
 
   history: {
+    api: typeof window !== 'undefined' ? WINDOW().history : {
```
Vic-Dev: Remove this, it's not used

packages/@storefront/flux-capacitor/src/core/actions/creators.ts
```diff
@@ -176,13 +185,13 @@ namespace ActionCreators {
   /**
    * Makes a request for the details for a given product.
    * @param  {string} id - The id for a specific product.
-   * @param {boolean} redirect - Indicates whether fetch is a result of a single product redirect.
+   * @param {boolean} buildAndParse - Indicates whether fetch is a result of a single product redirect.
```
Vic-Dev: Maybe leave this as `redirect`

packages/@storefront/flux-capacitor/src/core/actions/payloads.ts
```diff
@@ -2,17 +2,34 @@ import Configuration from '../configuration';
...
+      url?: string;
+      route?: string;
+      shouldFetch?: boolean;
+    }
+  }
+
   export namespace Fetch {
     export interface Override {
       request?: any;
+      redirect?: boolean;
```
Vic-Dev: Put this back in details

packages/@storefront/flux-capacitor/src/core/actions/payloads.ts
```diff
@@ -2,17 +2,34 @@ import Configuration from '../configuration';
...
+      buildAndParse?: boolean;
```
Vic-Dev: Maybe don't need

packages/@storefront/flux-capacitor/src/core/configuration.ts
```diff
@@ -73,6 +73,19 @@ interface Configuration {
    * personalization configuration
    */
   personalization?: Configuration.Personalization;
+
+  /**
```
Vic-Dev: Remove, not using this

packages/@storefront/flux-capacitor/src/core/selectors.ts
```diff
@@ -5,6 +5,35 @@ import Search, { MAX_RECORDS } from './adapters/search';
 import Store from './store';
 
 namespace Selectors {
+  /**
+   * Returns the history configuration.
+   */
+  export const historyConfig = (state: Store.State) =>
```
Vic-Dev: Remove. Not used

packages/@storefront/core/src/services/url.ts
```diff
@@ -24,166 +20,98 @@ class UrlService extends BaseService<UrlService.Options> {
...
       }
     }
   }
 
-  buildUrlAndReplaceHistory = ({ state, route }: { state: Store.State, route: string }) => {
-    const url = this.beautifier.build(route, this.urlState[route](state));
-    this.replaceHistory(url);
+  replaceState() {
+    if (this.opts.history && typeof this.opts.history.replaceState === 'function') {
+      return (data, title, url) => this.opts.history.replaceState(data, title, url);
```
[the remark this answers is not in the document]
Vic-Dev: We do only calculate it on initialization (if you look at line 37 when we call `initHistory` with `replaceState` and `pushState` we call these functions, so they only get the return values. For this one we can just return `this.opts.history.replaceState` if that's what you meant, and on line 55

packages/@storefront/flux-capacitor/src/core/actions/payloads.ts
Vic-Dev: Seems important

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
Vic-Dev: Probably fine as is

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
Vic-Dev: there's no other `console` in flux I just searched for it

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Added
+- SF-1185: Make a URL store
+  - `initHistory`, `pushState`, `updateHistory`, and `refreshState` methods on `flux`
+  - `updateHistory` action creator that triggers history `pushState` and `replaceState`, and updates the `history` section in the store
+  - `data.present.history` store section
+    - eg, ```
```
jmbtutor: The start of the code fence should be on a separate line for it to display properly. You can also add `js` to mark it as JavaScript.

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,22 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Added
+- SF-1185: Make a URL store
+  - `initHistory`, `pushState`, `updateHistory`, and `refreshState` methods on `flux`
```
jmbtutor: Full sentences, please.

packages/@storefront/flux-capacitor/src/core/observer.ts
```diff
@@ -107,6 +108,38 @@ namespace Observer {
...
         present: {
           // tslint:disable-next-line max-line-length
+          history: ((emitSearchUpdated: Observer, emitDetailsUpdated: Observer, emitPastPurchaseUpdated: Observer, emitCustomRouteUpdated: Observer, emitUrlUpdated: Observer, emitRouteUpdated: Observer) =>
+            (oldState: Store.History, newState: Store.History, path: string) => {
+              if (oldState !== newState) {
+                if (oldState.url !== newState.url) {
+                  emitUrlUpdated(oldState.url, newState.url, path);
+
+                  if (newState.shouldFetch) {
+                    if (newState.route === Routes.SEARCH) {
```
jmbtutor: `switch` on `newState.route`.

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
```diff
@@ -175,6 +190,7 @@ export namespace Middleware {
     ];
     const middleware = [
       ...normalizingMiddleware,
+      Middleware.updateHistory(flux),
       Middleware.saveStateAnalyzer(),
       Middleware.injectStateIntoRehydrate,
       Middleware.validator,
```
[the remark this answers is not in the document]
jmbtutor: If it's related and it directly affects the change, do it here. Otherwise, do it separately so we can revert it separately if we need to. Maybe switch it first so it gets absorbed in the major release.

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
```diff
@@ -138,13 +153,13 @@ export namespace Middleware {
...
-      } else if (SAVE_STATE_ACTIONS.includes(action.type) && !hasDispatched) {
+      } else if (SAVE_STATE_ACTIONS.includes(action.type)) {
         hasDispatched = true;
-        dispatch({ type: Actions.SAVE_STATE });
       }
 
-      if (batchLevel === 0) {
+      if (batchLevel === 0 && hasDispatched) {
         hasDispatched = false;
+        dispatch({ type: Actions.SAVE_STATE });
```
jmbtutor: Flagging for later discussion. I'm not sure if this change is necessary or correct.

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
```diff
@@ -20,6 +20,24 @@ declare module 'redux' {
...
+
+export const STOREFRONT_APP_ID = 'GroupBy StoreFront';
+
+export interface History {
+  parse: Parse;
+  build: Build;
+  pushState: (data: any, title: string, url: string) => void;
+  replaceState: (data: any, title: string, url: string) => void;
+  initialUrl: string;
+  noop?: () => void;
```
jmbtutor: Why do we define a noop here?

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
jmbtutor: Actually, that makes sense because we don't assume our environment in flux. So have we been silently catching errors?

But also, I think this function now knows too much about its caller since it refers to `UrlService`. Is it even useful to catch here? I think it should be up to the caller to handle because a following `then` is meaningless after getting to the `catch`.

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
```diff
@@ -52,12 +74,60 @@ class FluxCapacitor extends Emitter {
...
+      replaceState,
+      parse: (u) => Promise
+        .resolve(parse(u))
+        .catch((e) => console.warn('UrlService parse promise failed', e)),
+    };
+
+    this.updateHistory({ shouldFetch: true, url: initialUrl, method: this.history.noop });
+  }
+
+  // this function is here to support legacy implementations >.<"
```
jmbtutor:
```js
/**
 * @deprecated
 */
```
says the same thing in a way that shows up in the docs, too.

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
```diff
@@ -52,12 +74,60 @@ class FluxCapacitor extends Emitter {
...
+          this.actions.updateHistory({
+            url,
+            route: urlState.route || route,
+            request: urlState.request || request,
+            shouldFetch: urlState.shouldFetch,
+            method: urlState.method
+          })
+        );
+      });
+    } catch (e) {
```
jmbtutor: You can't get here unless something in the `this.history.parse(url)` chain is undefined. `parse` returns a Promise, which has its own error handling.

packages/@storefront/flux-capacitor/test/unit/core/observer.ts
```diff
@@ -288,6 +288,84 @@ suite('Observer', ({ expect, spy, stub }) => {
...
+          expect(emit).to.be.calledWith(Events.CUSTOM_URL_UPDATED);
+        });
+
+        it('should not emit when urls differ, and shouldFetch is false', () => {
+          const oldState = { url: 'www.google.ca', route: utils.Routes.SEARCH, shouldFetch: false };
+          const newState = { url: 'www.example.ca', route: utils.Routes.DETAILS, shouldFetch: false };
+
+          observers.data.present.history(oldState, newState, path);
+
+          expect(emit).to.be.calledWith(Events.URL_UPDATED);
```
jmbtutor: Can we assume that `emit` is only called once? If so, we won't need the checks below.

packages/@storefront/flux-capacitor/test/unit/core/reducers/data/history.ts
```diff
@@ -0,0 +1,52 @@
...
+        payload: <any>newState
+      });
+
+      expect(reducer).to.eql({ ...state, ...newState });
+    });
+
+    it('should return state on default', () => {
+      const reducer = history(state, <any>{});
+
+      expect(reducer).to.eql(state);
```
jmbtutor: This is a pass-through, so this should be `eq`.

packages/@storefront/flux-capacitor/test/unit/core/utils.ts
```diff
@@ -59,4 +59,78 @@ suite('utils', ({ expect, spy, stub }) => {
       expect(utils.normalizeToFunction(o)(o2)).to.eql({ ...o2, ...o });
     });
   });
+
+  describe('filterState()', () => {
+    it('should filter config from state and remove products when history length is 0', () => {
+      const data = { a: 'b', past: [{ a: 'b' }], present: { products: [1,2,3,4,5] } };
+      const payload = { e: 'f' };
+      const fullPayload = { method: () => null, ...payload };
```
jmbtutor: Seems like `payload` semantically should come first.

packages/@storefront/flux-capacitor/test/unit/flux-capacitor.ts
```diff
@@ -99,29 +99,189 @@ suite('FluxCapacitor', ({ expect, spy, stub }) => {
...
+            initialUrl,
+            pushState: push,
+            replaceState: replace,
+            noop: flux.history.noop
+          });
+          resultParse(url).then((response) => expect(initialParse).to.be.calledWith(url));
+        });
+
+        it('should set up parse with warning on error', () => {
+          // TODO: set up throw test when switch from `console.warn` to `app.log.warn`
```
jmbtutor: Use `it.skip` so that it doesn't show up as passing.

packages/@storefront/flux-capacitor/test/unit/flux-capacitor.ts
```diff
@@ -99,29 +99,189 @@ suite('FluxCapacitor', ({ expect, spy, stub }) => {
...
+          flux.history.build = build;
+          flux.history.parse = stub().withArgs(url).resolves({ route, request });
+          flux.store = <any>{ dispatch, getState: () => store };
+          flux.actions = <any>{ updateHistory };
+
+          flux.updateHistory(<any>{ shouldFetch, method });
+        });
+
+        it('should handle url parse errors', () => {
+          // TODO: when switch away from `console.warn`
```
jmbtutor: `it.skip`

packages/@storefront/infinite-scroll/src/infinite-scroll/index.ts
```diff
@@ -365,7 +365,7 @@ class InfiniteScroll {
   };
 }
 
-interface InfiniteScroll extends Core.Tag<InfiniteScroll.Props, InfiniteScroll.State> {}
+interface InfiniteScroll extends Core.Tag<InfiniteScroll.Props, InfiniteScroll.State> { }
```
jmbtutor: Revert.

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
Vic-Dev: I'm p sure it's not needed. Will take it out.

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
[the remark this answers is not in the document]
Vic-Dev: Sure makes sense. I don't think it's directly related I just noticed it and was like 🤦‍♀️

packages/@storefront/core/src/services/url.ts
```diff
@@ -24,166 +20,96 @@ class UrlService extends BaseService<UrlService.Options> {
...
-    this.app.flux.on(Events.HISTORY_SAVE, this.updateHistory);
-    this.app.flux.on(Events.HISTORY_REPLACE, this.buildUrlAndReplaceHistory);
+  build(route: string, state: Store.State) {
+    return this.beautifier.build(route, this.urlState[route](state));
   }
 
-  updateHistory = ({ state, route }: { state: Store.State, route: string }) => {
-    const url = this.beautifier.build(route, this.urlState[route](state));
-
+  pushState() {
```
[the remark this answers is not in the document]
Vic-Dev: if it's a getter how will it access the correct `this` context?

packages/@storefront/core/src/services/url.ts
```diff
@@ -24,166 +20,96 @@ class UrlService extends BaseService<UrlService.Options> {
...
+          return WINDOW().history.pushState(data, title, url);
+        }
       }
     }
   }
 
-  buildUrlAndReplaceHistory = ({ state, route }: { state: Store.State, route: string }) => {
-    const url = this.beautifier.build(route, this.urlState[route](state));
-    this.replaceHistory(url);
+  replaceState() {
```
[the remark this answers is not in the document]
Vic-Dev: Same question on using a getter

packages/@storefront/core/src/services/url.ts
```diff
@@ -24,166 +20,96 @@ class UrlService extends BaseService<UrlService.Options> {
...
-      } catch (e) {
-        this.app.log.warn('unable to push state to browser history', e);
+      return (data, title, url) => {
+        const redirectFnResult = typeof this.opts.redirects === 'function'
+          ? this.opts.redirects(url)
+          : this.opts.redirects[url];
+        if (redirectFnResult) {
+          return WINDOW().location.assign(redirectFnResult);
+        } else {
+          return WINDOW().history.pushState(data, title, url);
```
jrmykolyn: @Vic-Dev Can we inject the `window.history` object as a default at instantiation, rather than accessing it within each instance method?

For example, the methods which currently access `WINDOW().history.<method>` would be updated to read `this.defaults.history.<method>` (or similar).

Vic-Dev: @jrmykolyn `history` methods have to be called on `window` iirc, so can you be more specific with what you're suggesting?

jrmykolyn: @Vic-Dev I believe that the history methods must be called in the context of the history object itself, or of an object that implements the same interface.

That aside, I should have said: what about something like this:

```
this.defaults.history = <implementation-provided-api> || WINDOW().history;
```

From what I can tell, this should still work.

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
Vic-Dev: Theoretically if someone wants to call these methods with a noop function they can use the one that's here. Idk I can take it out if you think it's not needed.

packages/@storefront/flux-capacitor/src/flux-capacitor.ts
Vic-Dev: This was kept in to preserve old behaviour. This is the warning we get when you return a `route` that isn't a known `route` (eg, run presets w/ just `http://localhost:8080/`)

packages/@storefront/flux-capacitor/src/core/actions/payloads.ts
```diff
@@ -2,9 +2,25 @@ import Configuration from '../configuration';
...
 
 namespace Payload {
+  export namespace History {
+    export interface State extends ProtoState {
+      method: Function;
+      request?: any;
+    }
+
+    export interface ProtoState {
+      // TODO: add in the future
```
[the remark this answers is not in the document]
Vic-Dev: Why do we need this if we already have a type definition? https://github.com/groupby/storefront/pull/28/files/d83d76ffaa7c8e38b7acd3d27aa11c0c3ca7dcdc#diff-df780f9863ab20aba10f58f77ab14c83R35 I'm wary about putting external links that could go out of date. The point is that calling the `pushState` method has to match the type definition that we say it matches, it doesn't necessarily have to do what the `window.history.pushState` method does, but it does have to have the right definition.
I'm going to take out the `title` stuff for now since it's not implemented, but yes it would be to call the `pushState` and `replaceState` methods with

packages/@storefront/flux-capacitor/test/unit/flux-capacitor.ts
```diff
@@ -99,29 +99,178 @@ suite('FluxCapacitor', ({ expect, spy, stub }) => {
...
+
+      describe('updateHistory()', () => {
+        it('should parse the url and dispatch updateHistory', (done) => {
+          const url = 'www.url.com';
+          const shouldFetch = true;
+          const method = () => null;
+          const FETCH = 'fetch';
+          const route = 'search';
+          const request = { a: 'b' };
+          const dispatch = spy(() => {
```
jmbtutor: This feels odd, but I'm not sure there's a better way to do it unless you return the promise from `flux.updateHistory`.

packages/@storefront/infinite-scroll/src/infinite-scroll/index.ts
```diff
@@ -353,7 +353,7 @@ class InfiniteScroll {
   };
 
   replaceState = () => {
-    if (!this.state.firstLoad) this.flux.replaceState(this.state.route);
+    if (!this.state.firstLoad) this.flux.replaceState(this.state.route, true);
```
jmbtutor: I'm surprised the linter didn't complain about this.

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,23 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Added
+- SF-1185: Make a URL store
+  - `initHistory`, `pushState`, `updateHistory`, and `refreshState` methods added on `flux`.
+  - `updateHistory` action creator that triggers history `pushState` and `replaceState`, and updates the `history` section in the store.
+  - `data.present.history` store section was created:
```
jmbtutor: Mention the new selectors as well.

packages/@storefront/flux-capacitor/src/core/actions/payloads.ts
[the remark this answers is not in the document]
jrmykolyn: Noted, withdrawn.

packages/@storefront/core/src/services/url.ts
jrmykolyn: This has been addressed by setting up the `history` API methods in the `UrlService` constructor. Resolving.

jrmykolyn:
My feedback has been addressed, approving.

packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [major]
+### Changed
+- SF-1185: Make a URL store
```
Vic-Dev:
```suggestion
- SF-1185: Make a URL store
```

packages/@storefront/flux-capacitor/src/core/observer.ts
```diff
@@ -107,6 +108,42 @@ namespace Observer {
       data: {
         present: {
           // tslint:disable-next-line max-line-length
+          history: ((emitSearchUpdated: Observer, emitDetailsUpdated: Observer, emitPastPurchaseUpdated: Observer, emitCustomRouteUpdated: Observer, emitUrlUpdated: Observer, emitRouteUpdated: Observer) =>
+            (oldState: Store.History, newState: Store.History, path: string) => {
+              if (oldState !== newState) {
```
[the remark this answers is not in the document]
Vic-Dev: Is this what you meant?
```suggestion
if (oldState === newState) {
                return;
              } else if (oldState.url === newState.url && oldState.route === newState.route) {
                return;
              }

              emitUrlUpdated(oldState.url, newState.url, path);

              if (newState.shouldFetch) {
                switch (newState.route) {
                  case Routes.SEARCH:
                    emitSearchUpdated(oldState, newState, path);
                    break;
                  case Routes.DETAILS:
                    emitDetailsUpdated(oldState, newState, path);
                    break;
                  case Routes.PAST_PURCHASE:
                    emitPastPurchaseUpdated(oldState, newState, path);
                    break;
                  default:
                    emitCustomRouteUpdated(oldState, newState, path);
                }
              }

              emitRouteUpdated(oldState.route, newState.route, path);
```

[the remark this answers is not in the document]
jmbtutor: Yes for this:
```js
if (oldState === newState) return;
```
but I don't think you can generalize the second case unless the route will always change with the URL and vice versa. If that's actually the case, then you can merge the two conditions and do this: `oldState === newState || oldState.url === newState.url && oldState.route === newState.route`. I don't think you should do that, though.

packages/@storefront/infinite-scroll/src/infinite-scroll/index.ts
Vic-Dev: Sorry I'm not sure what you're referring to. Needs `{}`?

packages/@storefront/flux-capacitor/test/unit/flux-capacitor.ts
Vic-Dev: Any thoughts on how to make this a better test?

packages/@storefront/infinite-scroll/src/infinite-scroll/index.ts
jmbtutor: I was expecting it to say that you should reformat it like so:
```js
if (!this.state.firstLoad) {
  this.flux.replaceState(this.state.route, true);
}
```

I'm personally okay with this, though, so I won't ask you to change it.

packages/@storefront/flux-capacitor/test/unit/flux-capacitor.ts
jmbtutor: Nope.  An awkward test usually means the thing being tested could be written better, but I'm not sure it's worth it.

packages/@storefront/core/CHANGELOG.md
jmbtutor: "history store"

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,24 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Added
+- SF-1185: Make a URL store
```
jmbtutor: "history store"

packages/@storefront/infinite-scroll/CHANGELOG.md
```diff
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [patch]
+### Changed
+- SF-1185: Make a URL store
```
jmbtutor: "history store"

PR #29, opened 2018-12-19
Add script to add an unreleased section to a changelog

jmbtutor:


scripts/add-unreleased-section.sh
```diff
@@ -0,0 +1,81 @@
...
+      ;;
+  esac
+done
+
+shift $((OPTIND - 1))
+
+cd "${BASH_SOURCE%/*}/.."
+
+package="$1"
+changelog="packages/@storefront/${1}/CHANGELOG.md"
```
JoeyDeol: You should use the package variable instead of ${1} when setting the changelog variable.

scripts/add-unreleased-section.sh
```diff
@@ -0,0 +1,81 @@
...
+
+  echo "ERROR:" "$@" >&2
+  exit $exit_code
+}
+
+print_usage() {
+  cat <<EOF
+Usage: ${0##*/} <package>
+       ${0##*/} -h
+Adds an unreleased section to the CHANGELOG.md for <package>.
```
jrmykolyn: Add note to the effect of: This script must be executed from the root of the `storefront` repository.

jmbtutor: Line 50 takes care of that, so it can be called from anywhere.

jrmykolyn: Noted, resolving.

PR #31, opened 2018-12-21
 SF-1200 View Previous Search Terms in SAYT

blevy115:
Made initial previous search terms component


packages/@storefront/sayt/src/previous-search-terms/index.html
```diff
@@ -0,0 +1,3 @@
+<gb-list items="{state.previousSearches}" item-alias="previous-search" item-props="{ onClick: props.onClick }">
+  <yield/>
```
jrmykolyn: This should probably have some kind of default output (ie. something *within* the `<yield />` tag).

packages/@storefront/sayt/src/previous-search-terms/index.ts
```diff
@@ -0,0 +1,37 @@
+import { provide, tag, Events, Selectors, Store, Tag } from '@storefront/core';
+import Sayt from '../sayt';
+
+@provide('previousSearch', (props) => props)
+@tag('gb-sayt-previous-search', require('./index.html'))
+class PreviousSearch {
+  props: PreviousSearch.Props = {
+    onClick: (query) => () => this.actions.search(query),
+  } as any;
```
jrmykolyn: Does this need to be cast as `any`? If not, lets remove.

packages/@storefront/sayt/src/previous-search-terms/index.ts
```diff
@@ -0,0 +1,37 @@
...
+
+  constructor() {
+    const previousSearches = []
+    this.state = {...this.state, previousSearches}
```
jrmykolyn: Since we're using `previousSearches` immediately, lets inline the assignment like so:

```
this.state = { previousSearches: [] };
```

Also, since this component doesn't extend an existing class, we don't need to spread `this.state`.

packages/@storefront/sayt/src/previous-search-terms/index.ts
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  init() {
+    this.flux.on(Events.ORIGINAL_QUERY_UPDATED, this.updatePreviousSearches);
+  }
+
+  updatePreviousSearches = (originalQuery: string) => {
```
jrmykolyn: Lets change this from a function expression to a method declaration. ie.

```
updatePreviousSearches(originalQuery: string) {
  ...
}
```

Note that you'll also need to bind the method when registering it as a callback within `init()`. ie.

```
init() {
    this.flux.on(Events.ORIGINAL_QUERY_UPDATED, this.updatePreviousSearches.bind(this));
}
```

packages/@storefront/sayt/src/previous-search-terms/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    if(!this.state.previousSearches.includes(originalQuery) && this.state.previousSearches.length < 6) {
+      this.state.previousSearches.push(originalQuery)
+    } else if (!this.state.previousSearches.includes(originalQuery) && this.state.previousSearches.length >= 6) {
+      this.state.previousSearches.shift()
+      this.state.previousSearches.push(originalQuery)
+    }
+  }
+}
+
+interface PreviousSearch extends Tag<PreviousSearch.Props, PreviousSearch.State> {}
```
jrmykolyn: The `PreviousSearch.State` interface has not been defined. Add it to the namespace below:

ie.

```
{
...
    export interface State {
        ...
    }
...
}
```

packages/@storefront/sayt/src/previous-search-terms/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    const previousSearches = []
+    this.state = {...this.state, previousSearches}
+  }
+
+  init() {
+    this.flux.on(Events.ORIGINAL_QUERY_UPDATED, this.updatePreviousSearches);
+  }
+
+  updatePreviousSearches = (originalQuery: string) => {
+    if(!this.state.previousSearches.includes(originalQuery) && this.state.previousSearches.length < 6) {
```
jrmykolyn: Lets split up the expressions into two separate conditionals:

ie.

```
IF ( does not include query )
    ..
  IF ( has less than max )
    ..
  ELSE
    ..
```

packages/@storefront/sayt/src/previous-search-terms/index.ts
```diff
@@ -0,0 +1,37 @@
...
+      this.state.previousSearches.push(originalQuery)
+    } else if (!this.state.previousSearches.includes(originalQuery) && this.state.previousSearches.length >= 6) {
```
jrmykolyn: The `6` magic number should be moved into the StoreFront config and given a default.

packages/@storefront/sayt/src/previous-search-terms/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    this.state = {...this.state, previousSearches}
+  }
+
+  init() {
+    this.flux.on(Events.ORIGINAL_QUERY_UPDATED, this.updatePreviousSearches);
+  }
+
+  updatePreviousSearches = (originalQuery: string) => {
+    if(!this.state.previousSearches.includes(originalQuery) && this.state.previousSearches.length < 6) {
+      this.state.previousSearches.push(originalQuery)
```
jrmykolyn: Consider doing this without mutation (ie. `[...previousSearches, originalQuery]`).

packages/@storefront/sayt/src/previous-search-terms/index.ts
```diff
@@ -0,0 +1,37 @@
...
+    } else if (!this.state.previousSearches.includes(originalQuery) && this.state.previousSearches.length >= 6) {
+      this.state.previousSearches.shift()
+      this.state.previousSearches.push(originalQuery)
```
jrmykolyn: Same as above re: refactoring this to an immutable equivalent.

packages/@storefront/sayt/src/previous-search-terms/index.html
```diff
@@ -0,0 +1,3 @@
+<gb-list items="{state.previousSearches}" item-alias="previous-search" item-props="{ onClick: props.onClick }">
```
jrmykolyn: The value provided to `item-props` is missing the additional set of braces. Change to:

```
item-props="{ { onClick: props.onClick } }"
```

packages/@storefront/sayt/src/previous-search-terms/index.ts
```diff
@@ -0,0 +1,37 @@
+import { provide, tag, Events, Selectors, Store, Tag } from '@storefront/core';
+import Sayt from '../sayt';
+
+@provide('previousSearch', (props) => props)
+@tag('gb-sayt-previous-search', require('./index.html'))
+class PreviousSearch {
+  props: PreviousSearch.Props = {
+    onClick: (query) => () => this.actions.search(query),
```
jrmykolyn: I think you have an extra function here. In this case, invoking `onClick` will return a function. In order to dispatch the `search` action, this second function would *also* need to be invoked.

jrmykolyn: I believe this should read:

```
onClick: (query) => this.actions.search(query),
```

packages/@storefront/sayt/src/previous-search-terms/index.ts
```diff
@@ -0,0 +1,37 @@
...
+      this.state.previousSearches.shift()
+      this.state.previousSearches.push(originalQuery)
+    }
+  }
+}
+
+interface PreviousSearch extends Tag<PreviousSearch.Props, PreviousSearch.State> {}
+namespace PreviousSearch {
+  export interface Props {
+    onClick: (query: string) => () => void;;
```
jrmykolyn: See above re: additional function.

jrmykolyn (changes requested):

packages/@storefront/sayt/src/previous-search-terms/index.ts
jrmykolyn: Additional `;`. Please remove.

jrmykolyn (changes requested):

packages/@storefront/sayt/src/previous-search-terms/index.ts
jmbtutor: Also, since this is TypeScript, you can simply define this outside like you do with `props`. No need for a `constructor()` function.

packages/@storefront/sayt/src/previous-search-terms/index.ts
jmbtutor: Instead of binding in the `on` call, which is not easily testable, do this instead:
```js
constructor() {
  this.updatePreviousSearches = this.updatePreviousSearches.bind(this);
}
```

You'll see us setting function expressions in other components, but we're planning to move to the `bind` approach in the future.

packages/@storefront/sayt/src/previous-search-terms/index.ts
jmbtutor: Also, please add a space after keywords (`if`, `for`, `while`, etc.).

jmbtutor (dismissed):
Some additional feedback.

Vic-Dev (changes requested):
Missing many things for this feature to be passable:
- Configuration defaults & settings to make the feature extensible
- Implementation on SAYT
- Tests

Please come see me or another SF member in person because this feature requires a decent amount to be added to the library, and has been sitting for a while. If you need this urgently for a client, you should come see someone to pair and expedite this feature.

packages/@storefront/sayt/src/previous-search-terms/index.ts
blevy115: Ya I was thinking of doing something like that but wasnt sure how

packages/@storefront/sayt/src/previous-search-terms/index.ts
Vic-Dev: @jrmykolyn it's using currying because we don't want the click handler to use the event we want it to use the list item value

packages/@storefront/sayt/src/sayt/index.ts
```diff
@@ -27,6 +29,7 @@ class Sayt {
     this.subscribe('sayt:hide', this.setInactive);
     this.subscribe(Events.URL_UPDATED, this.setInactive);
     this.subscribe(Events.FETCHING_SEARCH, this.setInactive);
+    this.subscribe('previous:show', this.setPreviousActive);
```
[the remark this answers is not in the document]
Vic-Dev: Review stopped here for reasons of I'll be picking up this ticket and carrying it through to completion

Vic-Dev (changes requested):

Vic-Dev (changes requested):
NOTE: may need to store the previousSearches in a cookie because many SF clients either go outside of SF in between pages, don't behave as a spa, or in other ways can lose the query state between transitions

packages/@storefront/sayt/test/unit/previous-search-terms.ts
```diff
@@ -0,0 +1,79 @@
...
+      handler = previousSearchTerms.updatePreviousSearches;
+    })
+
+    it('should add search term if not in array and array length less than limit', () => {
+      handler(query);
+      expect(previousSearchTerms.state.previousSearches).to.eql(['diamond rings'])
+    })
+
+    it('should not add the term if it already exist in the array', () => {
+      previousSearchTerms.state.previousSearches = ['diamond rings', 'saphire', 'rings', 'diamond', ]
```
Vic-Dev: do this instead:
```
const previousSearches = [query, 'saphire', 'rings', 'diamond'];
previousSearchTerms.state.previousSearches = previousSearches;
```

packages/@storefront/sayt/test/unit/previous-search-terms.ts
```diff
@@ -0,0 +1,79 @@
...
+      handler(query);
+      expect(previousSearchTerms.state.previousSearches).to.eql(['diamond rings', 'saphire', 'rings', 'diamond'])
+    })
+
+    it('should remove first term and add new search term if array at limit', () => {
+      previousSearchTerms.state.previousSearches = ['saphire', 'rings', 'opals', 'necklaces', 'bracelet', 'earrings'];
```
Vic-Dev:
```
const previousSearches = ['saphire', 'rings', 'opals', 'necklaces', 'bracelet', 'earrings'];
previousSearchTerms.state.previousSearches = previousSearches;
```

packages/@storefront/sayt/test/unit/previous-search-terms.ts
```diff
@@ -0,0 +1,79 @@
...
+      handler(query);
```
Vic-Dev: Add newline above `handler(query);` and below

packages/@storefront/sayt/test/unit/previous-search-terms.ts
```diff
@@ -0,0 +1,79 @@
...
+      expect(previousSearchTerms.state.previousSearches).to.eql(['rings', 'opals', 'necklaces', 'bracelet', 'earrings', 'diamond rings'])
```
Vic-Dev: `expect(previousSearchTerms.state.previousSearches).to.eql([...previousSearches.slice(1), query])`

packages/@storefront/sayt/test/unit/previous-search-terms.ts
```diff
@@ -0,0 +1,79 @@
...
+    })
+
+    it('should move search term to beginning of list if in list and searched again', () => {
+      previousSearchTerms.state.previousSearches = ['saphire', 'rings', 'opals', 'necklaces', 'diamond rings', 'earrings'];
```
Vic-Dev:
```
const previousSearches = ['saphire', 'rings', 'opals', 'necklaces', query, 'earrings'];
previousSearchTerms.state.previousSearches = previousSearches;
```

packages/@storefront/sayt/test/unit/previous-search-terms.ts
```diff
@@ -0,0 +1,79 @@
...
+      handler(query);
```
Vic-Dev: Add newline above `handler(query);` and below

packages/@storefront/sayt/test/unit/previous-search-terms.ts
```diff
@@ -0,0 +1,79 @@
...
+      expect(previousSearchTerms.state.previousSearches).to.eql(['diamond rings', 'saphire', 'rings', 'opals', 'necklaces', 'earrings'])
```
Vic-Dev: `expect(previousSearchTerms.state.previousSearches).to.eql([query, ...previousSearches.slice(0, 4), ...previousSearches.slice(5)]);`

Vic-Dev (changes requested):
Finished off the review. Also needs to implement using a cookie instead of local state, which we can discuss irl

Vic-Dev (changes requested):
For implementing the cookie:
- Should live in the `autocomplete` service

PR #33, opened 2018-12-27
SF-1237: Add support for sorting past purchase products

jrmykolyn:


packages/@storefront/flux-capacitor/src/core/requests/utils.ts
```diff
@@ -59,10 +59,14 @@ namespace RequestHelpers {
       query: Selectors.pastPurchaseQuery(state),
       refinements: Selectors.pastPurchaseSelectedRefinements(state),
       skip: Selectors.pastPurchasePageSize(state) * (Selectors.pastPurchasePage(state) - 1),
-      sort: RequestAdapter.extractSort(Selectors.pastPurchaseSortSelected(state)),
     };
 
-    return <Request>{ ...request, ...overrideRequest };
+    const sort = RequestAdapter.extractPastPurchaseSort(
+      <any>overrideRequest.sort || Selectors.pastPurchaseSortSelected(state),
```
JoeyDeol: Why is this being cast to type of any?

packages/@storefront/flux-capacitor/test/unit/core/requests/utils.ts
```diff
@@ -202,9 +204,44 @@ suite('requests helpers', ({ expect, stub, spy }) => {
...
+
+    SORT_FIELDS.forEach((field) => {
+      it(`should handle the "${field}" sort`, () => {
+        ppSortStub.returns({ field });
+
+        const req = RequestHelpers.pastPurchaseProducts(state);
+
+        expect(req.sort).to.eql({ type: 'ByIds', ids: [1, 2, 3] });
+
+        ppSortStub.returns(sort);
```
JoeyDeol: ppSortStub seems to be only used inside of this "it" block, should we consider moving it down here instead of having it in the BeforeEach() if it isn't being used anywhere else?

packages/@storefront/core/test/unit/services/urlUtils.ts
```diff
@@ -673,5 +673,72 @@ suite('URL Service', ({ expect, spy, stub }) => {
...
+
+    describe('mergePastPurchaseSortsState()', () => {
+      it('should call `mergeSortsState()` with `Selectors.pastPurchaseSort`', () => {
+        const pastPurchaseSort = { foo: 'bar' };
+        const mergeSortsStateStub = stub(Utils, 'mergeSortsState');
+        stub(Selectors, 'pastPurchaseSort').returns(pastPurchaseSort);
+
+        Utils.mergePastPurchaseSortsState(<any>{}, <any>{});
+
+        expect(mergeSortsStateStub).to.be.calledWith(pastPurchaseSort);
```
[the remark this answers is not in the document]
jrmykolyn: Good call, will implement.

packages/@storefront/core/test/unit/services/urlUtils.ts
```diff
@@ -673,5 +673,72 @@ suite('URL Service', ({ expect, spy, stub }) => {
...
+
+    describe('mergeSearchSortsState()', () => {
+      it('should call `mergeSortsState()` with `Selectors.sorts`', () => {
+        const sort = { foo: 'bar' };
+        const mergeSortsStateStub = stub(Utils, 'mergeSortsState');
+        stub(Selectors, 'sorts').returns(sort);
+
+        Utils.mergeSearchSortsState(<any>{}, <any>{});
+
+        expect(mergeSortsStateStub).to.be.calledWith(sort);
```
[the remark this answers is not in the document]
jrmykolyn: Good call, will implement.

packages/@storefront/flux-capacitor/src/core/requests/utils.ts
jrmykolyn: In this case, the `sort` property is cast to `any` because the `overrideRequest` partially implements the `Request` interface (exposed by the `groupby-api` package). This interface says that the `sort` property can be either a 'field' sort, or a 'by id' sort. However, StoreFront expects the sort to be of the 'field'-type, which has a different shape.

See below for the interface definition:
https://github.com/groupby/api-javascript/blob/develop/src/models/request.ts#L13

packages/@storefront/flux-capacitor/test/unit/core/requests/utils.ts
JoeyDeol: Ah I see, understood. Resolving feedback.

packages/@storefront/flux-capacitor/src/core/requests/utils.ts
JoeyDeol: As per our IRL discussion 2018/12/27: I think that an update to what StoreFront expects in this case is required, however that update is outside of the scope of this ticket. Potentially we should resolve this down the line in another ticket. Also as a note: you mentioned that a potential implementation for this already exists on a branch, we should pull that in when we end up tackling this issue.

packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
```
jmbtutor: `patch`?

packages/@storefront/flux-capacitor/src/core/adapters/past-purchases.ts
```diff
@@ -33,6 +34,14 @@ namespace PastPurchasesAdapter {
     return [...skus].sort(({ [field]: lhs }, { [field]: rhs }) => rhs - lhs);
   };
 
+  export const sortSkusByType = (skus: Store.PastPurchases.PastPurchaseProduct[], type: number = -1) => {
```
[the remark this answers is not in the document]
jrmykolyn: Implemented, resolving.

packages/@storefront/flux-capacitor/src/core/requests/utils.ts
jrmykolyn: This update has been described in SF-1244. Resolving.

packages/@storefront/flux-capacitor/src/core/adapters/past-purchases.ts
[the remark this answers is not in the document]
jrmykolyn: Implemented, resolving.

packages/@storefront/flux-capacitor/test/unit/core/adapters/past-purchases.ts
```diff
@@ -121,4 +121,40 @@ suite('PastPurchase Adapter', ({ expect, stub }) => {
       });
     });
   });
+
+  describe('sortSkusByType()', () => {
+    let skus;
```
Vic-Dev: `const skus = [{ quantity: 1, lastPurchased: 420 }];`

packages/@storefront/flux-capacitor/test/unit/core/adapters/past-purchases.ts
```diff
@@ -121,4 +121,40 @@ suite('PastPurchase Adapter', ({ expect, stub }) => {
...
+    let sortSkusStub;
+
+    beforeEach(() => {
+      skus = [{ quantity: 1, lastPurchased: 420 }];
```
Vic-Dev: remove

packages/@storefront/flux-capacitor/src/core/adapters/past-purchases.ts
```diff
@@ -33,6 +34,14 @@ namespace PastPurchasesAdapter {
     return [...skus].sort(({ [field]: lhs }, { [field]: rhs }) => rhs - lhs);
   };
 
+  export const sortSkusByType = (skus: Store.PastPurchases.PastPurchaseProduct[], type?: number) => {
```
Vic-Dev: Why is type optional? It's literally called sort skus by type and also never called without type except for in the test testing it can be called without type. pls

packages/@storefront/flux-capacitor/src/core/reducers/data/past-purchases.ts
```diff
@@ -43,17 +49,17 @@ export const DEFAULTS: State = <any>{
   query: '',
   sort: {
```
Vic-Dev: I think you should add to the configuration adapter file https://github.com/groupby/storefront/blob/master/packages/%40storefront/flux-capacitor/src/core/adapters/configuration.ts#L33-L38
so that if someone wants to re-order their sort (or add additional sorts, like by price), they can do so and we will automatically set the types properly based on whether the given sort is in `SORT_FIELDS`

packages/@storefront/flux-capacitor/src/core/requests/utils.ts
```diff
@@ -59,10 +59,16 @@ namespace RequestHelpers {
...
+      Selectors.pastPurchases(state)
+    );
+
+    // Since the final past purchase sort is extracted from the override request,
```
Vic-Dev: Instead of doing this, why don't you do something like this:
```
const { sort, ...override } = overrideRequest;
```

Then you can put the sort in request and spread `override` after `request`.

packages/@storefront/flux-capacitor/test/unit/core/adapters/past-purchases.ts
jrmykolyn: Noted, will implement.

packages/@storefront/flux-capacitor/test/unit/core/adapters/past-purchases.ts
jrmykolyn: Noted, will implement.

packages/@storefront/flux-capacitor/src/core/adapters/past-purchases.ts
jrmykolyn: As per @jmbtutor:

```
I think type should be optional and left as undefined if not given. That way, it can't accidentally collide with an enum value.
```

How do we resolve?

packages/@storefront/flux-capacitor/src/core/reducers/data/past-purchases.ts
jrmykolyn: I believe that update is described in the following ticket: https://groupby.atlassian.net/browse/SF-1240.

Can we leave that as a separate item of work?

packages/@storefront/flux-capacitor/src/core/requests/utils.ts
jrmykolyn: Implemented, resolving.

packages/@storefront/flux-capacitor/src/core/adapters/past-purchases.ts
Vic-Dev: @jmbtutor can you explain your reasoning? I think using a optional parameter here before it's needed is over-engineering this problem. In any case, I don't feel strongly about it so it can be included.

packages/@storefront/flux-capacitor/src/core/reducers/data/past-purchases.ts
Vic-Dev: No. Being able to configure sorts and setting up sorts are the same implementation. For example, when sort was set up for search the configuration was done as the same ticket. It's one chunk of work in my opinion, and this feature is incomplete without it

packages/@storefront/flux-capacitor/src/core/reducers/data/past-purchases.ts
```diff
@@ -43,17 +49,17 @@ export const DEFAULTS: State = <any>{
...
     items: [{
-      field: 'Default',
+      field: SORT_FIELDS[SORT_ENUMS.DEFAULT],
       descending: true,
       type: SORT_ENUMS.DEFAULT,
```
Vic-Dev: Don't need both `type` and `field` as unique identifiers, just use `field`.

packages/@storefront/flux-capacitor/src/core/reducers/data/past-purchases.ts
```diff
@@ -31,6 +31,12 @@ export enum SORT_ENUMS {
   DEFAULT, MOST_PURCHASED, MOST_RECENT
 }
 
+export const SORT_FIELDS = [
+  'Default',
```
Vic-Dev: Store each string as a const or something and export the whole thing at the top level (maybe put it in an object like PAST_PURCHASE_SORTS so you can export less things)

jrmykolyn: Noted, will implement.

packages/@storefront/flux-capacitor/src/core/adapters/past-purchases.ts
jrmykolyn: Discussed IRL. Will leave in place.

packages/@storefront/flux-capacitor/src/core/reducers/data/past-purchases.ts
jrmykolyn: Noted. Will drop `type` and associated enum values.

packages/@storefront/flux-capacitor/src/core/reducers/data/past-purchases.ts
jrmykolyn: Addressed, resolving.

packages/@storefront/flux-capacitor/src/core/reducers/data/past-purchases.ts
jrmykolyn: Addressed, resolving.

packages/@storefront/flux-capacitor/src/core/reducers/data/past-purchases.ts
jrmykolyn: Updated implementation to allow implementations to specific past purchase sort options. Resolving.

packages/@storefront/flux-capacitor/src/core/reducers/data/past-purchases.ts
```diff
@@ -27,8 +27,15 @@ export type Action = Actions.ReceivePastPurchaseSkus
   | Actions.ReceivePastPurchaseTemplate;
 export type State = Store.PastPurchase;
 
-export enum SORT_ENUMS {
-  DEFAULT, MOST_PURCHASED, MOST_RECENT
+export namespace PAST_PURCHASE_SORTS {
+  export const DEFAULT = 'Default';
+  export const MOST_PURCHASED = 'Most Purchased';
+  export const MOST_RECENT = 'Most Recent';
+  export const ALL = [
```
JoeyDeol: Might want to potentially add some sort of leading identifier so people know that this is a collection of the other properties, but I am fine if it says as is. Thoughts @Vic-Dev @jmbtutor?

jrmykolyn: Dropped the `ALL` key entirely. Resolving.

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [patch]
+### Fixed
+- SF-1237: Fix issue where past purchase sorts are not correctly applied/transmitted.
+
```
Vic-Dev:
```
### Added
- SF-1237: Past purchase sort configuration
  - `PAST_PURCHASE_SORTS` exported
    - `DEFAULT` sorts past purchases based on the past purchase endpoint
    - `MOST_PURCHASED` sorts past purchases based on the quantity purchased
    - `MOST_RECENT` sorts past purchases based on the latest purchased timestamp
  - eg, {
      recommendations: {
        pastPurchases: {
          sort: {
            options: [{ field: PAST_PURCHASE_SORTS.MOST_RECENT, descending: true }, { field: PAST_PURHCASE_SORTS. MOST_PURCHASED, descending: true }],
            default: { field: PAST_PURCHASE_SORTS.MOST_RECENT, descending: true }
        }
      }
    }
```
p sure that's correct.

packages/@storefront/flux-capacitor/src/core/configuration.ts
```diff
@@ -337,6 +337,14 @@ namespace Configuration {
        */
       securedPayload: Recommendations.SecuredPayload | (() => Recommendations.SecuredPayload);
 
+      sort?: Configuration.ValueOptions<{
```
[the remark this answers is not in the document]
jrmykolyn: Create new interface under `Configuration` namespace to address this.

packages/@storefront/flux-capacitor/src/core/adapters/configuration.ts
```diff
@@ -123,8 +124,14 @@ namespace ConfigurationAdapter {
...
-  export const extractSorts = (config: Configuration): Store.SelectableList<Store.Sort> => {
-    const state = config.search.sort;
+  export const extractSearchSorts = (config: Configuration): Store.SelectableList<Store.Sort> =>
+    ConfigurationAdapter.extractSorts(config.search.sort);
+
+  export const extractPastPurchaseSorts = (config: Configuration): Store.SelectableList<Store.Sort> =>
+    ConfigurationAdapter.extractSorts(config.recommendations.pastPurchases.sort);
+
+  // tslint:disable-next-line max-line-length
+  export const extractSorts = (state: Configuration.ValueOptions<{ field: string; descending?: boolean; }> ): Store.SelectableList<Store.Sort> => {
```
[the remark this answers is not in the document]
jrmykolyn: Create new interface under `Configuration` namespace to address this.

packages/@storefront/flux-capacitor/CHANGELOG.md
Vic-Dev: youcan fixed the words some

packages/@storefront/flux-capacitor/test/unit/core/requests/utils.ts
```diff
@@ -202,9 +204,44 @@ suite('requests helpers', ({ expect, stub, spy }) => {
...
+
+    Object.keys(PAST_PURCHASE_SORTS).map((k) => PAST_PURCHASE_SORTS[k]).forEach((field) => {
+      it(`should handle the "${field}" sort`, () => {
+        pastPurchaseSortSelected.returns({ field });
+
+        const req = RequestHelpers.pastPurchaseProducts(state);
+
+        expect(req.sort).to.eql({ type: 'ByIds', ids: [1, 2, 3] });
+
+        pastPurchaseSortSelected.returns(sort);
```
jmbtutor: I don't think we need this.

packages/@storefront/flux-capacitor/CHANGELOG.md
jmbtutor: But make sure to actually format the object as a code block if you're going to include it.

jmbtutor: Also, CRAFDS, so make sure to put `Added` before `Fixed`.

jrmykolyn: Implemented.

packages/@storefront/flux-capacitor/test/unit/core/requests/utils.ts
jrmykolyn: Noted, will remove.

PR #38, opened 2019-01-02
SF-1226: Re-implement linter

jrmykolyn:


package.json
```diff
@@ -24,6 +24,14 @@
     "bundle:prod": "cross-env NODE_ENV=production webpack --config webpack.config.js",
     "docs": "typedoc --readme none ."
   },
+  "prettier": {
```
Vic-Dev: Why did you leave the other `prettier` configurations in the package.jsons?

jmbtutor: I don't remember, but it looks like it'll be fine without it. I'll remove them.

PR #50, opened 2019-01-14
SF-1082: Filter additional state before saving to browser history

jrmykolyn:


packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [patch]
+### Changed
+- Update `filterState` to remove navigations, template, and a subset of
+  autocomplete data when `history.length === 0`.
```
jmbtutor: I'm having second thoughts about the wrapping.

Vic-Dev: Should specify the configuration, ie `when history configuration is set to length: 0` or something

packages/@storefront/core/src/services/url.ts
```diff
@@ -157,15 +157,38 @@ class UrlService extends BaseService<UrlService.Options> {
...
+    let { navigations, products, template } = data.present;
+    let {
+      navigations: autocompleteNavigations,
+      products: autocompleteProducts,
+      template: autocompleteTemplate,
+    } = data.present.autocomplete;
+
+    if (this.app.config.history.length === 0) {
+      autocompleteNavigations = [];
+      autocompleteProducts = products = [];
```
Vic-Dev: I know we do this in the tests, but afaik we general don't in the code https://github.com/airbnb/javascript#variables--no-chain-assignment

packages/@storefront/core/src/services/url.ts
```diff
@@ -157,15 +157,38 @@ class UrlService extends BaseService<UrlService.Options> {
...
+      autocompleteTemplate = template = <any>{};
```
Vic-Dev: same thing here (no-chain-assignment)

packages/@storefront/core/src/services/url.ts
jrmykolyn: Sounds good, will remove.

packages/@storefront/core/src/services/url.ts
jrmykolyn: Ditto above.

packages/@storefront/core/CHANGELOG.md
jrmykolyn: Since the other CHANGELOG entries do not wrap, I will revert this change.

jrmykolyn: Updated. Resolving.

packages/@storefront/core/test/unit/services/url.ts
```diff
@@ -454,8 +454,21 @@ suite('URL Service', ({ expect, spy, stub, itShouldBeCore, itShouldExtendBaseSer
   });
 
   describe('filterState()', () => {
-    it('should filter config from state and remove products when history length is 0', () => {
-      const data = { a: 'b', past: [{ a: 'b' }], present: { products: [1,2,3,4,5] } };
+    it('should filter config from state and remove products, navigations, templates, and autocomplete data when history length is 0', () => {
```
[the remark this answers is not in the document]
jrmykolyn: Addressed. Resolving.

packages/@storefront/core/src/services/url.ts
```diff
@@ -157,15 +157,40 @@ class UrlService extends BaseService<UrlService.Options> {
 
   filterState(state: Store.State) {
     const { session: { config, ...session }, data, ...rootConfig } = state;
-    const products = this.app.config.history.length === 0 ? [] : data.present.products;
+    let { navigations, products, template } = data.present;
+    let {
+      navigations: autocompleteNavigations,
+      products: autocompleteProducts,
+      template: autocompleteTemplate,
+    } = data.present.autocomplete;
```
jmbtutor: I think you can actually do this:
```js
    let {
      navigations,
      products,
      template
      autocomplete: {
        navigations: autocompleteNavigations,
        products: autocompleteProducts,
        template: autocompleteTemplate,
      }
    } = data.present;
```

Up to you if you don't think this is clearer.

packages/@storefront/core/src/services/url.ts
```diff
@@ -157,15 +157,40 @@ class UrlService extends BaseService<UrlService.Options> {
...
+
+    if (this.app.config.history.length === 0) {
+      autocompleteNavigations = [];
+      autocompleteProducts = [];
+      autocompleteTemplate = <any>{};
+      navigations = { allIds: [], byId: {}, sort: [] };
+      products = [];
+      template = <any>{};
```
jmbtutor: Will TypeScript complain if you remove the cast?

jrmykolyn: Yes, since original assignment is based on the state object, TypeScript expects the `template` properties to exist. Resolving.

packages/@storefront/core/src/services/url.ts
jrmykolyn: Looks good, will implement.

PR #56, opened 2019-01-16
SF-948: Update `autocomplete` to support debounced product requests

jrmykolyn:


packages/@storefront/core/test/unit/core/utils.ts
```diff
@@ -27,6 +27,76 @@ suite('utils', ({ expect, spy, stub }) => {
...
+
+    it('should return a function', () => {
+      expect(utils.debounce(() => null)).to.be.a('function');
+    });
+
+    it('should debounce by the value provided', () => {
+      utils.debounce(fn, 500)();
+      clock.tick(500);
+
+      // TODO: See if the 'fake timers' API allows assertions to be made about the mocked methods themselves.
```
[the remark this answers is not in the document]
jrmykolyn: Addressed, resolving.

packages/@storefront/core/test/unit/core/utils.ts
```diff
@@ -27,6 +27,76 @@ suite('utils', ({ expect, spy, stub }) => {
...
+      expect(fn).to.be.called;
+    });
+
+    it('should debounce by 0ms if no value is provided', () => {
+      utils.debounce(fn)();
+      clock.tick(0);
+
+      // TODO: See if the 'fake timers' API allows assertions to be made about the mocked methods themselves.
```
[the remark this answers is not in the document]
jrmykolyn: Addressed, resolving.

packages/@storefront/flux-capacitor/src/core/configuration.ts
```diff
@@ -214,6 +214,10 @@ namespace Configuration {
      * enable updating query when hovering over sayt elements
      */
     hoverAutoFill?: boolean;
+    /**
+     * specify optional debounce value in milliseconds
+     */
+    debounce?: number;
```
JoeyDeol: Q: Which debounce value takes precedence? Let's say a user sets a value in the configuration and directly invokes the debounce function through the utils, will there be some sort of clash.

packages/@storefront/sayt/src/autocomplete/index.ts
```diff
@@ -128,6 +144,8 @@ class Autocomplete {
   }
 }
 
+type TargetData = { query: string, refinement: string, field: string, pastPurchase: string };
```
JoeyDeol: Should this type be included into the Autocomplete namespace?

packages/@storefront/flux-capacitor/src/core/configuration.ts
jrmykolyn: **TL;DR**
A. The presence or absence of the `debounce` configuration value will have no effect on an implementations use of `utils.debounce()`.

A: This file specifies that a value for `debounce` *may* be provided as part of the StoreFront configuration object. If the value _is_ provided, then the `Autocomplete` component will use it ti debounce requests for new products. However, this value is not automatically applied to all invocations of `utils.debounce()`.

packages/@storefront/sayt/src/autocomplete/index.ts
```diff
@@ -100,19 +106,29 @@ class Autocomplete {
     if (activate) {
       this.state.selected = index;
       if (indexExists) {
-        this.updateProducts(target, updateQuery);
+        const data = this.parseTarget(target);
+        if (updateQuery) this.updateQuery(data.query);
+        this.updateProducts(data);
```
jmbtutor: Is there a sane way to reduce the nesting here or collapse it in some way?

packages/@storefront/sayt/src/autocomplete/index.ts
jrmykolyn: Good question. Since this type definition is only applied to methods of the `Autocomplete` class itself, which themselves are not particularly relevant in any other context, I would say no. That said, let's see if this is flagged as part of any other review?

packages/@storefront/sayt/test/unit/autocomplete.ts
```diff
@@ -356,27 +385,54 @@ suite('Autocomplete', ({ expect, spy, stub, itShouldProvideAlias }) => {
...
+
+      expect(parseTarget).to.be.calledWithExactly(target);
+    });
+
+    it('should call updateQuery with the query string', () => {
+      const index = 0;
+      const target = { classList: { add: () => null } };
+      const query = 'foo';
+      const updateQuery = (autocomplete.updateQuery = spy());
+      autocomplete.parseTarget = spy(() => ({ query }));
```
jmbtutor: Doesn't need to be a `spy` if there are no assertions on it.

packages/@storefront/sayt/test/unit/autocomplete.ts
```diff
@@ -109,15 +120,33 @@ suite('Autocomplete', ({ expect, spy, stub, itShouldProvideAlias }) => {
...
+    it('should debounce the updateProducts method', () => {
+      const fn = () => null;
+      const delay = 420;
+      const debounce = stub(utils, 'debounce');
+      autocomplete.updateProducts = fn;
+      select.withArgs(Selectors.config).returns({ autocomplete: { debounce: delay } });
+
+      autocomplete.init();
+
+      expect(debounce).to.be.calledWithExactly(fn, 420, autocomplete);
```
jmbtutor: Use `delay` instead of `420`.

packages/@storefront/sayt/test/unit/autocomplete.ts
```diff
@@ -109,15 +120,33 @@ suite('Autocomplete', ({ expect, spy, stub, itShouldProvideAlias }) => {
...
+    });
+
+    it('should bind the context of the updateProducts method', () => {
+        const bind = spy();
```
jmbtutor: Fix indentation.

packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,19 @@ All notable changes to this project will be documented in this file.
...
+### Added
+- SF-948: Expose `debounce` utility function.
+  - `debounce` may be used to decouple function invocation from execution.
+  - `debounce` returns a new function, and may be used as follows:
+
+  ```js
+  import { utils } from '@storefront/core';
+
+  const fn = () => 'foo';
+  const debouncedFn = utils.debounce(fn, 500);
```
jmbtutor: Show a usage of `debouncedFn`.

packages/@storefront/flux-capacitor/src/core/configuration.ts
jmbtutor: This name signals a boolean value to me. Maybe call it `debounceDelay` or `debounceThreshold`. Remember to change the changelog.

JoeyDeol: Sounds good, resolving this issue.

packages/@storefront/sayt/src/autocomplete/index.ts
JoeyDeol: Fair I will leave this up and see if it flagged by anyone else, but for my part I understand and agree for leaving it as is.

packages/@storefront/core/test/unit/core/utils.ts
```diff
@@ -27,6 +27,77 @@ suite('utils', ({ expect, spy, stub }) => {
...
+
+    beforeEach(() => {
+      clock = sinon.useFakeTimers();
+      setTimeout = spy((...args) => clock.setTimeout(...args));
+      clearTimeout = spy((...args) => clock.clearTimeout(...args));
+      fn = spy();
+      win = stub(utils, 'WINDOW').returns({ setTimeout, clearTimeout });
+    });
+
+    it('should return a function', () => {
```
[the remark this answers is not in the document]
jrmykolyn: My rationale is that the fact `debounce()` returns a new function is a core piece of functionality, and is worth describing in its own test case. You can read this as redundant, or, if you limit yourself to the test case descriptions, a more accurate overview of the intended behaviour.

That said, I'm happy to remove this if you'd like?

packages/@storefront/flux-capacitor/src/core/configuration.ts
jrmykolyn: I was under the impression that we follow the practice of prefixing our boolean values with `is/has/should/etc.`? In any case, I think your suggestions *are* better than just `debounce`. Will update.

packages/@storefront/sayt/src/autocomplete/index.ts
jrmykolyn: Good idea. Will revisit.

packages/@storefront/sayt/test/unit/autocomplete.ts
jrmykolyn: Whoops, good call. Will update.

packages/@storefront/sayt/test/unit/autocomplete.ts
jrmykolyn: Will update.

packages/@storefront/sayt/test/unit/autocomplete.ts
jrmykolyn: Good call. Will update.

packages/@storefront/core/test/unit/core/utils.ts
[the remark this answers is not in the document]
jmbtutor: I think we implicitly test in similar cases, but I will leave this to your judgement. I don't think it's necessarily wrong to have this test here.

packages/@storefront/sayt/src/autocomplete/index.ts
jrmykolyn: CC: @jmbtutor?

packages/@storefront/core/test/unit/core/utils.ts
```diff
@@ -27,6 +27,77 @@ suite('utils', ({ expect, spy, stub }) => {
...
+  describe('debounce', () => {
+    let clock;
+    let fn;
+    let win;
+    let setTimeout;
+    let clearTimeout;
+
+    beforeEach(() => {
+      clock = sinon.useFakeTimers();
+      setTimeout = spy((...args) => clock.setTimeout(...args));
```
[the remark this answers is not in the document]
jrmykolyn: Addressed. Resolving.

packages/@storefront/core/test/unit/core/utils.ts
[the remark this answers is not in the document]
jrmykolyn: Leaving this in place. Resolving.

packages/@storefront/sayt/test/unit/autocomplete.ts
jrmykolyn: Addressed. Resolving.

packages/@storefront/sayt/test/unit/autocomplete.ts
jrmykolyn: Addressed. Resolving.

packages/@storefront/sayt/test/unit/autocomplete.ts
jrmykolyn: Addressed. Resolving.

packages/@storefront/core/CHANGELOG.md
jrmykolyn: Addressed. Resolving.

packages/@storefront/flux-capacitor/src/core/configuration.ts
jrmykolyn: Changed to `debounceThreshold `. Resolving.

packages/@storefront/sayt/src/autocomplete/index.ts
jmbtutor: I think we should, in part because we do for similar cases where we require an extra type. Also, if an implementation were to extend this component, they may want to refer to it as well.

jrmykolyn: Noted. Will update.

CC: @JoeyDeol

packages/@storefront/sayt/src/autocomplete/index.ts
jrmykolyn: As discussed, there does not appear to be a way to refactor this without resulting in scope creep.

One of the aggravating factors is the manual addition/removal of a given CSS class. This logic may be removed if the component is updated to track both the 'selected' and 'currently hovered' elements within its own state. However, this update should be completed as part of a new ticket.

packages/@storefront/sayt/src/autocomplete/index.ts
jrmykolyn: Updated. Resolving.

packages/@storefront/sayt/CHANGELOG.md
```diff
@@ -4,6 +4,21 @@ All notable changes to this project will be documented in this file.
...
+## [Unreleased] [minor]
+### Changed
+- SF-948: Update `autocomplete` to support a debounced variation of `updateProducts()`.
+  - To apply the debounce, update the `autocomplete` configuration as follows:
+
+  ```js
+  ...
+  autocomplete: {
+    ...
+    debounce: 200, // Milliseconds by which the function should be debounced.
```
jmbtutor: Update this to reflect the new name change.

PR #60, opened 2019-01-18
SF-1264: Navigation component should not render hidden content

Vic-Dev:
I had to modify `navigationHeader` a bit because I noticed that the icon wasn't updating when the `shouldUpdate` mixin is enabled.

### DIFF
--- a/packages/@storefront/navigation/src/navigation-header/index.ts
+++ b/packages/@storefront/navigation/src/navigation-header/index.ts
@@ -9,15 +9,15 @@ class NavigationHeader {
   };
 
   init() {
-    this.setIcon(this.toggleIcon());
+    this.setIcon();
   }
 
   onUpdate() {
-    this.setIcon(this.toggleIcon());
+    this.setIcon();
   }
 
-  setIcon(icon: string) {
-    this.state = { icon };
+  setIcon() {
+    this.state = { icon: this.toggleIcon() };
   }
 
   toggleIcon() {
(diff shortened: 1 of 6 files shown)


packages/@storefront/navigation/src/navigation-header/index.ts
```diff
@@ -8,13 +8,25 @@ class NavigationHeader {
...
+  init() {
+    this.setIcon();
+  }
+
+  onUpdate() {
+    this.setIcon();
+  }
+
+  setIcon() {
+    this.state = { icon: this.toggleIcon() };
```
jmbtutor: This should be spreading `state` in case an implementation extends this class or we decide to add more properties to `state`.

packages/@storefront/navigation/test/unit/navigation-header.ts
```diff
@@ -15,4 +15,54 @@ suite('NavigationHeader', ({ expect, spy, stub, itShouldProvideAlias }) => {
...
+      expect(navigationHeader.state.icon).to.eq(icon);
+    });
+  });
+
+  describe('toggleIcon()', () => {
+    const toggleOpen = 'open';
+    const toggleClosed = 'closed';
+    const icons = { toggleOpen, toggleClosed };
+
+    it('should return the toggleOpen state when isActive is true', () => {
```
jmbtutor: "icon" instead of "state"

packages/@storefront/navigation/test/unit/navigation-header.ts
```diff
@@ -15,4 +15,54 @@ suite('NavigationHeader', ({ expect, spy, stub, itShouldProvideAlias }) => {
...
+      navigationHeader.props = <any>{ isActive: true, icons };
+
+      expect(navigationHeader.toggleIcon()).to.eq(toggleOpen);
+    });
+
+    it('should return the toggleClosed state when isActive is false', () => {
```
jmbtutor: "icon" instead of "state"

packages/@storefront/navigation/src/navigation-header/index.ts
Vic-Dev: good call 👍

PR #62, opened 2019-01-22
SF-1256: Add support for updating sort options after instantiation

jrmykolyn:


packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,70 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Changed
+- SF-1256: Persist sort and past purchase sort labels in the application store.
+- SF-1256: Update `search` and `pastPurchases` sort configurations to support labels.
+  - To apply sort labels via the `search` sort configuration;
```
jmbtutor: colon

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,70 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Changed
+- SF-1256: Persist sort and past purchase sort labels in the application store.
```
jmbtutor: We should consider standardizing our tenses in the changelog. Changelogs are traditionally in the past tense. I'm not going to ask you to change this, if you don't want to, though.

packages/@storefront/flux-capacitor/src/core/actions/payloads.ts
```diff
@@ -157,6 +157,12 @@ namespace Payload {
     isFetchingForward?: boolean;
     isFetchingBackward?: boolean;
   }
+
+  export interface Sort {
+    labels?: string[];
+    options: { field: string, descending?: boolean }[];
```
jmbtutor: The object type should be given a name and made into its own interface.

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
```diff
@@ -181,3 +181,40 @@ export const isNotFetching: Validator<boolean> = {
...
   msg: 'is already fetching'
 };
+
+export const hasValidLabels: Validator<Actions.Payload.Sort> = {
+  func: ({ labels }) => {
+    return !labels
+      || (
+        Array.isArray(labels)
+        && !!labels.length
+        && labels.every((label) => typeof label === 'string')
```
jmbtutor: If you use `some`, you won't have to check that the array is not empty. Up to you if you don't think it's clear.

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
```diff
@@ -181,3 +181,40 @@ export const isNotFetching: Validator<boolean> = {
...
+  msg: 'if present, labels must be an array of strings',
+};
+
+export const hasValidOptions: Validator<Actions.Payload.Sort> = {
+  func: ({ options }) => {
+    return (
+      Array.isArray(options)
+      && !!options.length
+      && options.every((option) => typeof option.field === 'string'
+        && (typeof option.descending === 'boolean' || !('descending' in option))
```
jmbtutor: As a matter of style, I would put the check for the presence of `'descending'` first.

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
```diff
@@ -181,3 +181,40 @@ export const isNotFetching: Validator<boolean> = {
...
+  },
+  msg: 'if present, labels must be an array of strings',
+};
+
+export const hasValidOptions: Validator<Actions.Payload.Sort> = {
+  func: ({ options }) => {
+    return (
+      Array.isArray(options)
+      && !!options.length
+      && options.every((option) => typeof option.field === 'string'
```
jmbtutor: If you have to break the conditional into multiple lines, I'd put the first condition on its own line as well.

packages/@storefront/flux-capacitor/test/unit/core/actions/validators.ts
```diff
@@ -652,4 +652,111 @@ suite('validators', ({ expect, spy, stub }) => {
...
+
+    [
+      'foo',
+      1,
+      true,
+      {},
+      () => null,
+      undefined,
+    ].forEach((val) => {
+      it(`should be invalid if options is a ${typeof val}`, () => {
```
jmbtutor: If you say "is of type" instead of "is a", you won't find yourself in the awkward situation where you get "is a undefined". Ditto for the same below.

packages/@storefront/sort/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [patch]
+### Changed
+- SF-1256: Sort component will now whenever it recieves new props.
```
jmbtutor: "will now" what?

packages/@storefront/sort/src/sort/index.ts
```diff
@@ -34,21 +34,35 @@ class Sort {
...
+      ...this.state,
+      sorts: this.extractSorts(),
+    };
+  }
+
   updateSorts = () => this.set({ sorts: this.extractSorts() });
 
   extractSorts() {
     let sorts;
+    let labels;
```
jmbtutor: Is it more legible to have this default to `sorts.labels`, then conditionally set `labels` inside the switch?

packages/@storefront/sort/src/sort/index.ts
```diff
@@ -34,21 +34,35 @@ class Sort {
...
     switch (this.props.storeSection) {
       case StoreSections.PAST_PURCHASES:
         sorts = this.select(Selectors.pastPurchaseSort);
+        labels = this.props.pastPurchasesLabels && this.props.pastPurchasesLabels.length
```
jmbtutor: `this.props.pastPurchaseLabels` and `this.props.labels` seem like good candidate for destructuring into local constants.

packages/@storefront/sort/test/unit/sort.ts
```diff
@@ -76,6 +76,19 @@ suite('Sort', ({ expect, spy, stub, itShouldBeConfigurable, itShouldProvideAlias
...
 
+  describe('onUpdate()', () => {
+    it('should extract and spread sorts into state', () => {
+      const state: any = sort.state = <any>{ a: 'b' };
+      const sorts = { c: 'd' };
+      const extractSorts = (sort.extractSorts = spy(() => sorts));
+
+      sort.onUpdate();
+
+      expect(extractSorts).to.be.called;
```
jmbtutor: This is assertion is probably redundant. You can't get `sort.state` to have the same `sorts` value if `extractSorts` isn't called.

packages/@storefront/flux-capacitor/src/core/store/index.ts
```diff
@@ -55,7 +55,7 @@ namespace Store {
     area: string;
     query: Query; // mixed
 
-    sorts: SelectableList<Sort>;
+    sorts: LabelSelectableList<Sort>;
```
jmbtutor: Maybe `SelectableListWithLabels` or `LabeledSelectableList`. `Label-` doesn't really tell me much about what the object is.

jrmykolyn:
Additional feedback received following demo:
- Update action creators to allow `[]` for `labels` field.
- Update reducers to use `[]` as default value if `labels` not provided.
- Update reducers to use `0` as default value if `selected` not provided.

CC: @jmbtutor @JoeyDeol

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
jrmykolyn: Agreed, will update.

packages/@storefront/flux-capacitor/src/core/actions/payloads.ts
jrmykolyn: I believe this interface actually exists elsewhere within `FluxCapacitor`. Will find and apply.

packages/@storefront/flux-capacitor/CHANGELOG.md
jrmykolyn: Agreed, will change.

packages/@storefront/flux-capacitor/CHANGELOG.md
jrmykolyn: Whoops, will update.

packages/@storefront/flux-capacitor/test/unit/core/actions/validators.ts
jrmykolyn: Good point, will update.

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
jrmykolyn: Noted, will apply.

packages/@storefront/sort/CHANGELOG.md
jrmykolyn: Whoops, looks like we missed this. Will update.

packages/@storefront/sort/src/sort/index.ts
jrmykolyn: Will investigate.

packages/@storefront/sort/src/sort/index.ts
jrmykolyn: Will investigate.

packages/@storefront/sort/test/unit/sort.ts
jrmykolyn: Noted, will remove.

packages/@storefront/flux-capacitor/src/core/store/index.ts
jrmykolyn: Good call, will update.

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
jrmykolyn: I'll revisit this as part of the 'allow empty arrays' feedback (provided during demo).

jrmykolyn:
Added `Do Not Merge` label while demo feedback is being implemented.

packages/@storefront/sort/src/sort/index.ts
jrmykolyn: This feedback item may have been addressed by the introduction of `extractLabels()`. Please re-review.

packages/@storefront/sort/src/sort/index.ts
jrmykolyn: This feedback item may have been addressed by the introduction of `extractLabels()`. Please re-review.

packages/@storefront/sort/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [patch]
+### Changed
+- SF-1256: Sort component will now update whenever it recieves new props.
```
jmbtutor: The component always updates when it receives new props. Be more specific.

jrmykolyn:
Completed second demo. See below for additional feedback:
- User should should always be able to select a newly applied sort option, even if shares an index with the previously selected option.

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
```diff
@@ -104,9 +104,15 @@ export const isPastPurchaseRefinementSelectedByIndex: Validator<Actions.Payload.
...
-  msg: 'past purchases sort is already selected'
+export const isPastPurchasesSortValid: Validator<number> = {
+  func: (index, state) => {
+    return (
+      typeof index === 'number'
+      && index >= 0
+      && index <= (Selectors.pastPurchaseSort(state).items.length - 1)
+    );
+  },
+  msg: 'past purchases sort index is not a number or is not within the sorts array'
```
jmbtutor: Perhaps "... is greater than the length of the sorts array", or a la Java, something something "out of bounds". The idea here is that the index won't be found as a value in the array as the text suggests, it's an index of the array.

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
```diff
@@ -181,3 +193,40 @@ export const isNotFetching: Validator<boolean> = {
...
+export const hasValidSelected: Validator<Actions.Payload.Sort> = {
+  func: (payload) => {
+    return !('selected' in payload)
+      || (
+        typeof payload.selected === 'number'
+        && payload.selected >= 0
+        && payload.selected <= payload.options.length - 1
+      );
+  },
+  msg: 'if present, must be a number between 0 and `options.length - 1`',
```
jmbtutor: I would say "is an index of the options array". If you want, you can give the range from 0 to `options.length - 1`, but use the actual calculated value.

packages/@storefront/flux-capacitor/test/unit/core/actions/validators.ts
```diff
@@ -652,4 +664,111 @@ suite('validators', ({ expect, spy, stub }) => {
...
+    });
+
+    it('should be invalid if selected is less than 0', () => {
+      expect(validators.hasValidSelected.func(<any>{
+        options: ['a', 'b', 'c'],
+        selected: -1,
+      })).to.be.false;
+    });
+
+    it('should be invalid if selected is greater than options.length - 1', () => {
```
jmbtutor: The description is precise, but maybe a bit too precise. Consider something like "past the end of the options array".

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
```diff
@@ -145,9 +151,15 @@ export const isCollectionDeselected: Validator<string> = {
...
-  msg: 'sort is already selected'
+export const isSortValid: Validator<number> = {
+  func: (index, state) => {
+    return (
+      typeof index === 'number'
+      && index >= 0
+      && index <= (Selectors.sorts(state).items.length - 1)
+    );
+  },
+  msg: 'sort index is not a number or is not within the sorts array'
```
jmbtutor: Ditto on the message.

packages/@storefront/flux-capacitor/test/unit/core/actions/validators.ts
JoeyDeol: Done, resolving feedback.

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
JoeyDeol: Done, resolving feedback.

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
JoeyDeol: Done, resolving feedback.

packages/@storefront/flux-capacitor/src/core/actions/validators.ts
JoeyDeol: Done, resolving feedback.

JoeyDeol:
2019/01/24 Demo'd this to @jmbtutor and demo was given a pass.

PR #63, opened 2019-01-22
SF-1268: Make @uiState() decorator to preserve UI state on unmount

Vic-Dev:


packages/@storefront/core/src/core/tag/decorators.ts
```diff
@@ -81,3 +82,33 @@ export function origin(name: string) {
...
+
+      if (onBeforeMount) {
+        return onBeforeMount.bind(this)(...args);
+      }
+    };
+
+    target.prototype.onBeforeUnmount = function(...args: any) {
+      this.actions.createComponentState(Tag.getMeta(this).name, this.props[prop], resolver(this.props, this.state));
+
+      if (onBeforeUnmount) {
```
jmbtutor: Ditto on the function check.

packages/@storefront/core/src/core/tag/decorators.ts
```diff
@@ -81,3 +82,33 @@ export function origin(name: string) {
...
   utils.setMetadata(target, CONFIGURABLE_KEY, true);
 }
+
+export function uiState<P extends object = any, S extends object = any>(
+  prop: string = 'uiValue',
+  resolver: (props: P, state: S) => any = (_, state) => state
+) {
+  return (target: TagConstructor) => {
+    const onBeforeMount = target.prototype.onBeforeMount;
+    const onBeforeUnmount = target.prototype.onBeforeUnmount;
```
jmbtutor: Optional, but you can reduce this to one line by destructuring.

packages/@storefront/core/src/core/tag/decorators.ts
```diff
@@ -81,3 +82,33 @@ export function origin(name: string) {
 export function configurable<P extends object>(target: TagConstructor) {
   utils.setMetadata(target, CONFIGURABLE_KEY, true);
 }
+
+export function uiState<P extends object = any, S extends object = any>(
+  prop: string = 'uiValue',
```
jmbtutor: Do we need this to be overridable? I would stick with one prop name for consistency. You can add it to the tag index file to add it for all tags (I think we do that for `storeSection`). I would also call it `uiKey` or something like that, because the value is being used as a key.

Vic-Dev:
> I'm not convinced that this is going to work reliably across multiple instances of the same class. I don't think that it should be up to the implementation to manage the ui state keys across the whole app, especially when they might not know all keys. Perhaps we can use a key based on its position in the DOM, perhaps using its XPath or CSS selector? With that, I think the approach would be to allow the ui key prop and fall back to the generated path if one is not provided.

I think the problem with using the CSS selector is that the DOM could potentially change. Still looking into the xpath implementation

packages/@storefront/core/src/core/tag/decorators.ts
```diff
@@ -81,3 +82,33 @@ export function origin(name: string) {
...
+    const onBeforeUnmount = target.prototype.onBeforeUnmount;
+
+    target.prototype.onBeforeMount = function(...args: any) {
+      const storedState = this.select(Selectors.uiTagState, Tag.getMeta(this).name, this.props[prop]);
+
+      if (storedState) {
+        this.state = { ...this.state, ...storedState };
+      }
+
+      if (onBeforeMount) {
```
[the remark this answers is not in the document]
Vic-Dev: I don't think this is necessary

packages/@storefront/core/src/core/tag/decorators.ts
```diff
@@ -81,3 +82,33 @@ export function origin(name: string) {
 export function configurable<P extends object>(target: TagConstructor) {
   utils.setMetadata(target, CONFIGURABLE_KEY, true);
 }
+
+export function uiState<P extends object = any, S extends object = any>(
+  prop: string = 'uiValue',
+  resolver: (props: P, state: S) => any = (_, state) => state
```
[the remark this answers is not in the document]
Vic-Dev: How would I access the alias in this context?

PR #66, opened 2019-01-29
SF-1248: Allow refinement-crumbs to update as required in order to have latest store data

JoeyDeol:


packages/@storefront/breadcrumbs/src/refinement-crumbs/index.ts
```diff
@@ -20,6 +20,10 @@ class RefinementCrumbs {
     this.updateState();
   }
 
+  shouldUpdate() {
```
jrmykolyn: Is there a way we can use the `shouldUpdate` params to determine whether or not an update is required?

ie. can we compare the indices of the currently `selected` refinements (for this navigation) with the previously selected indices?

packages/@storefront/breadcrumbs/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [patch]
+### Added
+- SF-1248: Added a `shouldUpdate` function to the `refinement-crumbs` component.
+  - The addition of this function allows for the component to update as needed and have the latest data from the application store.
```
jrmykolyn: I would include a note re: *why* this is necessary? For example: "Since the `field` prop does not change over the lifespan of the component, the `shouldUpdate` method is required in order to re-render in response to newly applied refinements."

packages/@storefront/breadcrumbs/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
...
+- SF-1248: Added `PAST_PURCHASE_NAVIGATIONS_UPDATED` event listener to `breadcrumbs` component `init()` function.
+  - This event listener was added so `refinement-crumbs` would update as needed in past purchase scenarios.
```
jrmykolyn: Add a note re: "parity with the 'search' listeners...".

packages/@storefront/breadcrumbs/CHANGELOG.md
JoeyDeol: Noted, will add that in.

packages/@storefront/breadcrumbs/CHANGELOG.md
JoeyDeol: Will do, resolving conversation.

JoeyDeol:
Closed PR because we have a different approach in mind now.

PR #71, opened 2019-02-04
Sf 1279: When hoverAutoFill is false, the search should be based on the term in the search box

Vic-Dev:


packages/@storefront/core/test/unit/services/autocomplete.ts
```diff
@@ -173,6 +178,19 @@ suite('Autocomplete Service', ({ expect, spy, stub, itShouldBeCore, itShouldExte
 
       expect(service.hasActiveSuggestion()).to.be.false;
     });
+
+    it('should return false if no autocomplete tags has an active selection, but hoverAutoFill is off', () => {
```
jrmykolyn: Should this test read "...if any autocomplete tags..."? If I understand correctly, we're asserting that `hasActiveSuggestion()` will return false in the case where 1 or more tags are active, but `hoverAutoFill` is disable?

packages/@storefront/core/src/services/autocomplete.ts
```diff
@@ -53,7 +53,9 @@ class AutocompleteService extends LazyService {
   }
 
   hasActiveSuggestion() {
-    return this.registeredAutocompleteTags.some((tag) => tag.isActive());
+    return this.registeredAutocompleteTags.some(
+      (tag) => tag.isActive() && tag.select(Selectors.config).autocomplete.hoverAutoFill
```
jmbtutor: `tag.select(Selectors.config).autocomplete.hoverAutoFill` should technically be the same for all iterations. Is there a way we can pull this out outside the `some`?

PR #74, opened 2019-02-13
SF-1275: Create configuration option to disable GbTracker beaconing

carloferrer:


carloferrer:
Ticket resolved via client's StoreFront config file.

PR #76, opened 2019-02-19
SF-1125: On details pages, allow parsing of collection/area via URL

carloferrer:


packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Added
+- SF-1125: Added option for parsing collection/area through URL on details pages.
```
jmbtutor:
```suggestion
- SF-1125: Added option for parsing `collection` and `area` through URL on details pages.
```

packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
...
+  - This new option can be leveraged via query string as follows: `?collection=Production&area=productsLeaf`
```
jmbtutor: Perhaps explain that `Production` and `productsLeaf` are examples of collection and area respectively.

packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
...
+  - Note: collection and area may be individually specified.
```
jmbtutor: No need for "Note:". I would also say "independently" instead of "individually'.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
```diff
@@ -18,6 +18,13 @@ export class DetailsUrlParser extends UrlParser<UrlBeautifier.DetailsUrlState> {
     const title = path.shift();
     const id = path.pop();
     let variants = [];
+    const queryParams: {[key: string]: string} = {};
```
jmbtutor: Padding.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
```diff
@@ -18,6 +18,13 @@ export class DetailsUrlParser extends UrlParser<UrlBeautifier.DetailsUrlState> {
...
+
+    if (Object.keys(uri.query).length) {
+      const filteredQueryParams = DetailsUrlParser.extractFilteredQueryParams(uri.query);
+      queryParams.area = filteredQueryParams.area;
+      queryParams.collection = filteredQueryParams.collection;
```
jmbtutor: These should be guarded to check for `undefined`. You don't want to add it if it doesn't exist.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
```diff
@@ -62,6 +69,12 @@ export class DetailsUrlParser extends UrlParser<UrlBeautifier.DetailsUrlState> {
 
     return variants;
   }
+
+  static extractFilteredQueryParams(queryParams: {[key: string]: string}) {
```
jmbtutor: Padding on the object type.

packages/@storefront/core/test/unit/core/url-beautifier/details/parser.ts
```diff
@@ -97,6 +97,23 @@ suite('DetailsUrlParser', ({ expect }) => {
     expect(parsed.variants).to.eql(expectedDetail.variants);
   });
 
+  // tslint:disable-next-line max-line-length
+  it('should extract query string parameters of a URL specifying "area" and "collection", and add them to the return object', () => {
```
jmbtutor: I would split this into two tests, one for area and one for collection.

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Changed
+- SF-1125: Details pages now leverage the new `RequestHelper`, `productDetails`.
+
+### Added
+- SF-1125: On details pages, requests are built via `productDetails`, a new `RequestHelper` that accounts for a specified collection and area.
```
jmbtutor: It's not clear that the new `RequestHelper` is what was added. State the addition first, then explain.

packages/@storefront/flux-capacitor/test/unit/core/requests/utils.ts
```diff
@@ -479,6 +479,39 @@ suite('requests helpers', ({ expect, stub, spy }) => {
...
+
+    const overrides = request;
+
+    beforeEach(() => {
+      stub(RequestHelpers, 'search').withArgs(state, overrides).returns(request);
+
+      stub(Selectors, 'history').withArgs(state).returns(history);
+      stub(Selectors, 'area').withArgs(state).returns(state.area);
+      stub(Selectors, 'collection').withArgs(state).returns(state.collection);
+    });
```
jmbtutor: I would put all of this in the test itself. They'll be factored out if necessary when new tests are made.

packages/@storefront/core/test/unit/core/url-beautifier/details/parser.ts
```diff
@@ -97,6 +97,23 @@ suite('DetailsUrlParser', ({ expect }) => {
...
+    const url = 'dress/293014?collection=joey&area=carlo';
+    const expectedDetail = {
+      variants: [],
+      collection: 'joey',
```
jrmykolyn: I think the use of your names for the area and collection values might be a little *too* fun. Consider using values which approximate real area and collection names.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
```diff
@@ -18,6 +18,13 @@ export class DetailsUrlParser extends UrlParser<UrlBeautifier.DetailsUrlState> {
     const title = path.shift();
     const id = path.pop();
     let variants = [];
+    const queryParams: {[key: string]: string} = {};
+
+    if (Object.keys(uri.query).length) {
+      const filteredQueryParams = DetailsUrlParser.extractFilteredQueryParams(uri.query);
```
jrmykolyn: This function feels a bit redundant. Is there a reason that we don't extract the `area` and `collection` properties directly from the `uri.query` object?

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
jrmykolyn: +1 for this.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
jrmykolyn: See above re: whether or not this function is actually necessary.

Alternatively, if you decide that it is necessary, it should have 1 or more accompanying tests.

packages/@storefront/flux-capacitor/test/unit/core/requests/utils.ts
```diff
@@ -479,6 +479,39 @@ suite('requests helpers', ({ expect, stub, spy }) => {
...
+
+    it('should create a product details request', () => {
```
jrmykolyn: Based on the mock data above, this should read something like: 'should create a product details request using `history`'. I think you need an accompanying test for cases where the `history` data is not set/does not exist.

packages/@storefront/core/test/unit/core/url-beautifier/details/parser.ts
JoeyDeol: lol will change.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
JoeyDeol: That makes sense, will remove unnecessary function.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
JoeyDeol: Removed unnecessary function.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
JoeyDeol: Agreed, will changed.

packages/@storefront/core/test/unit/core/url-beautifier/details/parser.ts
JoeyDeol: Will do.

packages/@storefront/flux-capacitor/test/unit/core/requests/utils.ts
JoeyDeol: Will add.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
```diff
@@ -18,6 +18,10 @@ export class DetailsUrlParser extends UrlParser<UrlBeautifier.DetailsUrlState> {
     const title = path.shift();
     const id = path.pop();
     let variants = [];
+    const queryParams: { [key: string]: string } = {
+      ...( uri.query.area ? { area: uri.query.area } : {} ),
+      ...( uri.query.collection ? { collection: uri.query.collection } : {} )
```
jrmykolyn: Sorry, I have to do: We don't pad the content inside parentheses.

packages/@storefront/core/CHANGELOG.md
```diff
@@ -6,9 +6,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased] [minor]
 ### Added
-- SF-1125: Added option for parsing collection/area through URL on details pages.
+- SF-1125: Added option for parsing `collection` and `area` through URL on details pages.
```
jmbtutor: Sorry I didn't catch this initially, but you should also mention that this is implemented in the default parser, and also outline what they need to include in their return value to be able to use this.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
```diff
@@ -18,6 +18,10 @@ export class DetailsUrlParser extends UrlParser<UrlBeautifier.DetailsUrlState> {
     const title = path.shift();
     const id = path.pop();
     let variants = [];
+    const queryParams: { [key: string]: string } = {
+      ...( uri.query.area ? { area: uri.query.area } : {} ),
```
jrmykolyn: I dawned on me that the query string parameters should probably be configurable, falling back to `area` and `collection` if the implementation doesn't provide custom keys. I think the PR should be updated to include this functionality.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
```diff
@@ -18,13 +18,10 @@ export class DetailsUrlParser extends UrlParser<UrlBeautifier.DetailsUrlState> {
...
-    const queryParams: {[key: string]: string} = {};
-
-    if (Object.keys(uri.query).length) {
-      const filteredQueryParams = DetailsUrlParser.extractFilteredQueryParams(uri.query);
-      queryParams.area = filteredQueryParams.area;
-      queryParams.collection = filteredQueryParams.collection;
-    }
+    const queryParams: { [key: string]: string } = {
+      ...( uri.query.area ? { area: uri.query.area } : {} ),
+      ...( uri.query.collection ? { collection: uri.query.collection } : {} )
```
jmbtutor: I'm not sure how I feel about this. I would have just done something like the following:

```js
const queryParams: { [key: string]: string } = {};
const { area, collection } = uri.query;

if (area) queryParams.area = area;
if (collection) queryParams.collection = collection;
```

But if you choose to keep the spreading, remove the padding between the `()`.

packages/@storefront/core/test/unit/core/url-beautifier/details/parser.ts
```diff
@@ -97,6 +97,36 @@ suite('DetailsUrlParser', ({ expect }) => {
     expect(parsed.variants).to.eql(expectedDetail.variants);
   });
 
+  it('should extract query string parameters of a URL specifying "area" and add it to the return object', () => {
```
jrmykolyn: The test cases above don't include the trailing "...and add it to the return object", so I think we can omit that text here.

Also, I think there's a singular/plural mismatch ("...query string parameters..." vs. "URL specifying 'area'...").  This should probably read something to the effect of:

'should extract and apply the 'area' query string parameter'

packages/@storefront/core/test/unit/core/url-beautifier/details/parser.ts
```diff
@@ -97,6 +97,36 @@ suite('DetailsUrlParser', ({ expect }) => {
...
+        id: '293014',
+        title: 'dress',
+      },
+      id: '293014',
+    };
+
+    expect(parser.parse(url)).to.eql(expectedDetail);
+  });
+
+  it('should extract query string parameters of a URL specifying "collection" and add it to the return object', () => {
```
jrmykolyn: Same as above re: messaging.

packages/@storefront/flux-capacitor/test/unit/core/requests/utils.ts
```diff
@@ -479,6 +479,38 @@ suite('requests helpers', ({ expect, stub, spy }) => {
     });
   });
 
+  describe('productDetails()', () => {
+    it('should create a product details request using history', () => {
+      const state: any = { area: 'oof', collection: 'rab', };
+      const request = { area: 'foo', collection: 'bar', };
+      const history = { request: { area: 'foo', collection: 'bar', } };
+      const overrides = request;
+
```
jrmykolyn: Since the 'act' and 'assert' steps have been collapsed into on, everything above the `expect()` invocation is technically part of the 'arrange' step. Lets remove the new lines between the `const` declaration and the `stub`, as well as between the `stub` function calls.

packages/@storefront/flux-capacitor/test/unit/core/requests/utils.ts
```diff
@@ -479,6 +479,38 @@ suite('requests helpers', ({ expect, stub, spy }) => {
...
+      expect(RequestHelpers.productDetails(state)).to.eql(request);
+    });
+
+    it('should create a product details request using store fallbacks if history is not available', () => {
+      const state: any = { area: 'oof', collection: 'rab', };
+      const request: any = state;
+      const history = { request: {} };
+      const overrides = state;
+
```
jrmykolyn: Same as above.

packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Added
+- SF-1125: Added option for parsing `collection` and `area` through URL on details pages.
+  - This new option can be leveraged via query string as follows: `?collection=Production&area=productsLeaf`
```
jrmykolyn: See below re: suggestion to make these parameters configurable.

packages/@storefront/core/src/core/url-beautifier/details/parser.ts
carloferrer: @jrmykolyn what do you mean by "configurable"?  Like, add an option in the config of whether or not to expect `area` and `collection` as query string params?

packages/@storefront/core/test/unit/core/url-beautifier/factory.ts
```diff
@@ -15,7 +15,7 @@ const URL_BEAUTIFIER: any = { a: 'b' };
 suite('BeautifierFactory', ({ expect, stub }) => {
   let config: any;
 
-  beforeEach(() => config = URL_BEAUTIFIER.config = {});
+  beforeEach(() => config = URL_BEAUTIFIER.config = { details: { params: {}}});
```
jmbtutor: Padding.

packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,41 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [minor]
+### Added
+- SF-1125: Added option for parsing `collection` and `area` through URL on details pages via configurable key-names.
+  - This new option can be leveraged via query string as follows: `?area=Production&collection=productsLeaf`
+    - Where `Production` and `productsLeaf` are examples of collection and area, respectively.
```
jrmykolyn: Add a note re: the default values (shown above). For example, something to the effect of:
- Where 'area' and 'collection' are the default key-names for `area` and `collection`, respectively.

packages/@storefront/core/test/unit/core/url-beautifier/factory.ts
JoeyDeol: Fixed.

packages/@storefront/core/CHANGELOG.md
JoeyDeol: Added.

jrmykolyn:
Nice work! Approved.

carloferrer:
Staging this commit message for when this PR gets approved...
> On details pages, allow parsing of collection and/or area (for finer control over a query for a product record) via URL query string parameters.
> - By default, 'area' and 'collection' will be parsed as the key-names for `area` and `collection`, respectively.
> - Custom key-names for parsing `area` and `collection` may be independently specified.
> - Parsing of `area` and `collection` may be independently disabled.

packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,43 @@ All notable changes to this project will be documented in this file.
...
+  - Collection and area may be independently specified.
+  - Area and collection may be referenced by custom key-names using the `details` property within the `beautifier` object.
+    - Conversely, parsing of area and collection may be independently disabled by providing `false` as the value for a custom key-name.
+  - Sample configuration:
+
+  ```js
+  ...
+  services :{
```
jmbtutor: The space is in the wrong spot.

PR #80, opened 2019-03-15
SF-1315: Allow users to set query property within the store to empty string through requests

JoeyDeol:


packages/@storefront/core/test/unit/services/urlUtils.ts
```diff
@@ -689,6 +689,60 @@ suite('URL Service', ({ expect, spy, stub }) => {
...
+
+        const newQueryState = Utils.mergeSearchQueryState(state, request);
+
+        expect(newQueryState).to.eql({
+          a: 'b',
+          original: 'foo'
+        });
+      });
+
+      it('should merge search query state properly when given a new request with a query property of empty string', () => {
```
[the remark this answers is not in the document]
carloferrer: 😮

packages/@storefront/core/test/unit/services/urlUtils.ts
```diff
@@ -689,6 +689,60 @@ suite('URL Service', ({ expect, spy, stub }) => {
       });
     });
 
+    describe.only('mergeSearchQuery', () => {
```
jmbtutor: Take off this `.only`.

packages/@storefront/core/src/services/urlUtils.ts
```diff
@@ -245,7 +245,7 @@ namespace UrlUtils {
 
   export const mergeSearchQueryState = (state: Store.State, request: UrlBeautifier.SearchUrlState) => ({
     ...state.data.present.query,
-    original: request.query || Selectors.query(state),
+    original: request.query === '' ? '' : request.query || Selectors.query(state),
```
jmbtutor: I would check for `== undefined` instead. Better to check for the fallback case than to check for a special case.

packages/@storefront/core/CHANGELOG.md
```diff
@@ -4,6 +4,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [patch]
+### Changed
+- SF-1315: Allow users to set `query` to empty string within the application store.
```
jmbtutor: You should mention something about this being in the URL service, specifically the parser.

packages/@storefront/core/test/unit/services/urlUtils.ts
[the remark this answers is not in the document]
JoeyDeol: Lint appeased.

packages/@storefront/core/CHANGELOG.md
JoeyDeol: Will do.

JoeyDeol:
Demo Passed.

JoeyDeol:
Leaving a Do Not Merge label here until we can pull this change into the BJs implementation in order to solve a bug revealed by the new angular changes. 

Note: a potential change will have to be made within the BJs project as well in order to ensure this release is used properly.

JoeyDeol:
Decided to release this anyways as it would be a bug for more users beyond BJs.

PR #81, opened 2019-03-20
SF-1317: gb-carousel now sets tiles to the carousel on page load

GregYaniw:
https://groupby.atlassian.net/browse/SF-1317

### DIFF
--- a/packages/@storefront/structure/src/carousel/index.ts
+++ b/packages/@storefront/structure/src/carousel/index.ts
@@ -28,7 +28,7 @@ class Carousel {
 
   onMount() {
     utils.WINDOW().addEventListener('resize', this.forceUpdate);
-
+    
     // forcing an update here instead of on init because 
     // the items have a width of 0 if items are simply set on init. 
     // this is done here to recalculate styles.
--- a/packages/@storefront/structure/src/carousel/index.ts
+++ b/packages/@storefront/structure/src/carousel/index.ts
@@ -26,6 +26,13 @@ class Carousel {
   };
   animationEndCallback: NodeJS.Timer = null;
 
+  init() {
+    this.state = {
+      ...this.state,
+      items: this.cloneItems(),
+    };
+  }
+
   onMount() {
     utils.WINDOW().addEventListener('resize', this.forceUpdate);
   }
--- a/packages/@storefront/structure/src/carousel/index.ts
+++ b/packages/@storefront/structure/src/carousel/index.ts
@@ -29,9 +29,7 @@ class Carousel {
   onMount() {
     utils.WINDOW().addEventListener('resize', this.forceUpdate);
 
-    // forcing an update here instead of on init because 
-    // the items have a width of 0 if items are simply set on init. 
-    // this is done here to recalculate styles.
+    // the items have a width of 0 if items are simply set on init, so do an update to recalculate styles
     this.set(true);
   }
 
--- a/packages/@storefront/structure/src/carousel/index.ts
+++ b/packages/@storefront/structure/src/carousel/index.ts
@@ -31,7 +31,9 @@ class Carousel {
 
     // this forces an update on mount which in turn sets
     // items to state from props by calling cloneItems
-    // so that items display within the carousel
+    // so that items display within the carousel. 
+    // We are doing this because on mount of gb-carousel, it is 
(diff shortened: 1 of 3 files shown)


packages/@storefront/structure/src/carousel/index.ts
```diff
@@ -28,6 +28,7 @@ class Carousel {
 
   onMount() {
     utils.WINDOW().addEventListener('resize', this.forceUpdate);
+    this.set(true);
```
jmbtutor: Add a brief comment here explaining why this is necessary.

packages/@storefront/structure/src/carousel/index.ts
```diff
@@ -28,6 +28,11 @@ class Carousel {
 
   onMount() {
     utils.WINDOW().addEventListener('resize', this.forceUpdate);
+
+    // this forces an update on mount which in turn sets
+    // items to state from props by calling cloneItems
+    // so that items display within the carousel
+    this.set(true);
```
JoeyDeol: Can you leave a comment as to why we need to use `this.set(true)` in the onMount? I think it's clear as to what it will do here, but more context on why it's here would be good.

packages/@storefront/structure/src/carousel/index.ts
```diff
@@ -28,6 +28,13 @@ class Carousel {
 
   onMount() {
     utils.WINDOW().addEventListener('resize', this.forceUpdate);
+
+    // this forces an update on mount which in turn sets
+    // items to state from props by calling cloneItems
+    // so that items display within the carousel. 
+    // We are doing this because on mount of gb-carousel, it is 
+    // not populating the carousel with the item data contained in state.
```
jmbtutor: The reason for doing `this.set(true)` here instead of simply setting the items in `init` is because styles need to be recalculated to take into account the number of items.

Consider something like this:

```js
// the items have a width of 0 if items are simply set on init, so do an update to recalculate styles
```

packages/@storefront/structure/src/carousel/index.ts
```diff
@@ -28,6 +28,9 @@ class Carousel {
 
   onMount() {
     utils.WINDOW().addEventListener('resize', this.forceUpdate);
+
+    // the items have a width of 0 if items are simply set on init, so do an update to recalculate styles
+    this.set(true);
```
jrmykolyn: @GregYaniw I would swap the `set()` call out for `this.forceUpdate()`, which more clearly expresses your intention. That said, I would still leave the comment in place.

packages/@storefront/structure/test/unit/carousel.ts
```diff
@@ -37,12 +37,22 @@ suite('Carousel', ({ expect, spy, stub, itShouldProvideAlias }) => {
...
+        carousel.set = () => null;
         stub(utils, 'WINDOW').returns({ addEventListener });
 
         carousel.onMount();
 
         expect(addEventListener).to.be.calledWithExactly('resize', carousel.forceUpdate);
       });
+
+      it('should force an update', () => {
+        const set = carousel.set = spy();
```
jrmykolyn: If you replace `set()` with `forceUpdate()`, these tests will have to be updated as well.

packages/@storefront/structure/src/carousel/index.ts
```diff
@@ -28,6 +28,9 @@ class Carousel {
 
   onMount() {
     utils.WINDOW().addEventListener('resize', this.forceUpdate);
+
+    // the items have a width of 0 if items are simply set on init, so do an update to recalculate styles
```
jrmykolyn: I understand that you need to force an update, but I don't quite get what you mean by "...if items are simply set on init..."? Can you expand?

packages/@storefront/structure/test/unit/carousel.ts
```diff
@@ -37,12 +37,22 @@ suite('Carousel', ({ expect, spy, stub, itShouldProvideAlias }) => {
     describe('onMount()', () => {
       it('should add event listener for window resize', () => {
         const addEventListener = spy();
+        carousel.set = () => null;
```
jrmykolyn: Since we swapped out `this.set()` for `this.forceUpdate()`, this override should also be updated.

PR #82, opened 2019-03-28
SF-1296: Reduce memory leaks

carloferrer:


packages/@storefront/core/src/core/tag/mixins/provide-consume.ts
```diff
@@ -34,10 +34,12 @@ export namespace ProvideConsume {
           this.one(Phase.UNMOUNT, () => {
             aliasTag.off(Phase.UPDATE, resetUpdated);
             aliasTag.off(Phase.UPDATED, softUpdateDependant);
+            this.off(Phase.UPDATE, markUpdated);
```
jrmykolyn: In this context, is `this` different from `aliasTag`? If not, is there a reason why we're not using `aliasTag`?

packages/@storefront/core/src/core/tag/index.ts
```diff
@@ -99,7 +98,10 @@ class Tag<P extends object = any, S extends object = any, A extends object = any
     }
   }
 
-  _removeEventHandlers = () => this._eventHandlers.forEach(([event, handler]) => this.flux.off(event, handler));
+  _removeEventHandlers = () => {
+    this._eventHandlers.forEach(([event, handler]) => this.flux.off(event, handler));
+    this._eventHandlers = [];
```
jmbtutor: Add comment about garbage collection here.

packages/@storefront/core/src/core/tag/mixins/provide-consume.ts
jmbtutor: Yes. There are two components in this context: `this`, which is the component that consumes the alias, and `aliasTag`, which is the component that provides the alias. Thinking about it, this line might not be necessary because Riot should be doing this cleanup on unmount anyway.

packages/@storefront/core/src/core/tag/mixins/provide-consume.ts
```diff
@@ -34,10 +34,11 @@ export namespace ProvideConsume {
           this.one(Phase.UNMOUNT, () => {
             aliasTag.off(Phase.UPDATE, resetUpdated);
             aliasTag.off(Phase.UPDATED, softUpdateDependant);
+            // allow alias tag to be garbage-collected
+            aliasTag = null;
```
jrmykolyn: @jmbtutor

To confirm: since the `aliasTag` object (and it's associated properties) will be garbage collected, the thinking here is that we don't need to explicit remove the `markUpdated` listener?

packages/@storefront/core/src/core/tag/mixins/provide-consume.ts
jrmykolyn: Noted. I have 1x follow question re: how the `aliasTag` is marked for garbage collection.

packages/@storefront/core/src/core/tag/mixins/provide-consume.ts
jmbtutor: No, it's that Riot unregisters all of its listeners for a given component when that component unmounts, so doing it ourselves is redundant.

jrmykolyn: Noted, thanks for clarifying.

carloferrer:
:pray:

PR #95, opened 2019-05-01
SF-1355: Suppress Redux actions following `location.assign()`

jrmykolyn:


packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [major|minor|patch]
```
jmbtutor: Fix the release type.

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
...
+### Added
+- SF-1355:
+  - Added `START_REDIRECT` and `DONE_REDIRECT` actions.
+  - Added `startRedirect()` and `stopRedirect()` action creators.
+- <Describe additions>
```
jmbtutor: Remove.

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased] [major|minor|patch]
+### Added
+- SF-1355:
```
jmbtutor: Distribute this into the two sub-points and promote the sub-points. The main point isn't the ticket, it's the things done.

packages/@storefront/flux-capacitor/CHANGELOG.md
```diff
@@ -4,6 +4,16 @@ All notable changes to this project will be documented in this file.
...
+  - Added `START_REDIRECT` and `DONE_REDIRECT` actions.
+  - Added `startRedirect()` and `stopRedirect()` action creators.
+- <Describe additions>
+
+### Fixed
+- SF-1355: Fixed an issue where navigating via redirect resulted in a 'dropped' browser history entry.
```
jmbtutor:
```suggestion
- SF-1355: Fixed an issue where navigating via redirect resulted in a "dropped" browser history entry.
```

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
```diff
@@ -147,6 +147,24 @@ export namespace Middleware {
...
+    let isRedirecting = false;
+
+    return () => (next) => (action) => {
+      if (action.type === Actions.START_REDIRECT) {
+        isRedirecting = true;
+      } else if (action.type === Actions.DONE_REDIRECT) {
+        isRedirecting = false;
+      }
+
+      if (isRedirecting) {
```
jmbtutor:
```js
if (!isRedirecting) return next(action);
```

Or in three lines if you think that reads better.

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
```diff
@@ -427,6 +432,48 @@ suite('Middleware', ({ expect, spy, stub }) => {
     });
   });
 
+  describe('redirectAnalyzer()', () => {
+    let next;
```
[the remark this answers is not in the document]
jmbtutor: `next` is already declared at the top level. No need to redefine it here.

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
```diff
@@ -427,6 +432,48 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+
+    beforeEach(() => {
+      next = sinon.spy();
```
jmbtutor: `next` is already assigned as a `spy()` in the top-level `beforeEach`, so this is redundant.

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
```diff
@@ -427,6 +432,48 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+    it('should not invoke the next middleware if the action is of type START_REDIRECT', () => {
+      const action = { type: 'START_REDIRECT' };
+
+      const result = redirectAnalyzer()(next)(action);
+
+      expect(result).to.be.undefined;
+      expect(next).not.to.be.called;
+    });
+
+    it('should not invoke the next middleware if the START_REDIRECT action has been dispatched previously', () => {
```
jmbtutor: Do you think it's worth it to test the complement of this (i.e. after a `DONE_REDIRECT`, make sure that subsequent actions pass through)?

packages/@storefront/flux-capacitor/src/core/actions/creators.ts
```diff
@@ -1176,6 +1176,23 @@ namespace ActionCreators {
...
     return createAction(Actions.START_APP, undefined);
   }
+
+  // navigation action creators
+  /**
+   * Fires the START_REDIRECT action.
+   * @return - Action with undefined.
+   */
+  export function startRedirect(): Actions.StartRedirect {
+    return createAction(Actions.START_REDIRECT, undefined);
```
JoeyDeol: Why are you returning `undefined` as the action's payload?

packages/@storefront/flux-capacitor/src/core/actions/creators.ts
```diff
@@ -1176,6 +1176,23 @@ namespace ActionCreators {
...
+  }
+
+  /**
+   * Fires the DONE_REDIRECT action.
+   * @return - Action with undefined.
+   */
+  export function stopRedirect(): Actions.DoneRedirect {
+    return createAction(Actions.DONE_REDIRECT, undefined);
```
JoeyDeol: Same as above^

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
JoeyDeol: Does this current approach mean that a `DONE_REDIRECT` action must be dispatched after a `START_REDIRECT` action in order for SF to return to normal functionality after a redirect?

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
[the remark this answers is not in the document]
jrmykolyn: Good flag, will remove.

packages/@storefront/flux-capacitor/src/core/actions/creators.ts
jrmykolyn: Currently, type declarations for actions that do not include a payload use one of two values: `undefined`; or `any` (see the links below for examples). In this case, I opted to use `undefined`.

https://github.com/groupby/storefront/blob/master/packages/%40storefront/flux-capacitor/src/core/actions/index.ts#L71

https://github.com/groupby/storefront/blob/master/packages/%40storefront/flux-capacitor/src/core/actions/index.ts#L237

packages/@storefront/flux-capacitor/src/core/actions/creators.ts
jrmykolyn: Response is same as above.

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
jrmykolyn: That's correct. `START_REDIRECT` causes all subsequent actions *other than* `DONE_REDIRECT` to be ignore. That said, `START_REDIRECT` is dispatched immediately after `window.location.assign()` is invoked. At some point, `assign()` *will* navigate to the target page. This will re-instantiate the application and restore the default behaviour.

packages/@storefront/flux-capacitor/CHANGELOG.md
jrmykolyn: Good catch, will do.

packages/@storefront/flux-capacitor/CHANGELOG.md
jrmykolyn: Good catch, will do.

packages/@storefront/flux-capacitor/CHANGELOG.md
jrmykolyn: Good point, will do.

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
jrmykolyn: Noted, will remove.

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
jrmykolyn: This is more concise, will implement.

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
jrmykolyn: @jmbtutor 

I actually went back-and-forth on this. My current thinking (and the reason that the corresponding test was omitted) is that the sole `DONE_REDIRECT` test actually accounts for this. Put another way, it doesn't actually matter if `START_REDIRECT` was dispatched prior to `DONE_REDIRECT`, so long as the latter ensures that subsequent actions are handled as normal.

In the spirit of 'completeness', I'm happy to add this test.

packages/@storefront/flux-capacitor/src/core/actions/creators.ts
JoeyDeol: Cool makes sense. Resolving this thread.

packages/@storefront/flux-capacitor/src/core/store/middleware.ts
JoeyDeol: Makes sense, resolving this thread.

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
jmbtutor: An implementation that calls `next` on `DONE_REDIRECT` but not set `isRedirecting` (preventing future actions from passing through) will still pass these tests, which is why I wondered if we should have it. Perhaps we can merge this new test with the existing `DONE_REDIRECT` test, since it'll only pass if `DONE_REDIRECT` works properly.

If we do that, perhaps the sole `START_REDIRECT` test is redundant for the same reason.

jrmykolyn: Those are good points. I'll consolidate the existing tests.

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
```diff
@@ -427,6 +432,30 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+      const startRedirectAction = { type: Actions.START_REDIRECT };
+      const doneRedirectAction = { type: Actions.DONE_REDIRECT };
+
+      redirectAnalyzer()(next)(startRedirectAction);
+      redirectAnalyzer()(next)({ type: 'FOO' });
+      redirectAnalyzer()(next)({ type: 'BAR' });
+      redirectAnalyzer()(next)({ type: 'BAZ' });
+      redirectAnalyzer()(next)(doneRedirectAction);
+
+      expect(next).to.have.callCount(1);
```
jmbtutor: Also assert which arguments were passed in.

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
```diff
@@ -427,6 +432,30 @@ suite('Middleware', ({ expect, spy, stub }) => {
...
+    // tslint:disable-next-line max-line-length
+    it('should suppress subsequent middleware upon receipt of the START_REDIRECT action, until the DONE_REDIRECT action is received', () => {
+      const startRedirectAction = { type: Actions.START_REDIRECT };
+      const doneRedirectAction = { type: Actions.DONE_REDIRECT };
+
+      redirectAnalyzer()(next)(startRedirectAction);
+      redirectAnalyzer()(next)({ type: 'FOO' });
+      redirectAnalyzer()(next)({ type: 'BAR' });
+      redirectAnalyzer()(next)({ type: 'BAZ' });
+      redirectAnalyzer()(next)(doneRedirectAction);
```
jmbtutor: Add one more action after the "done" action to test the "until the DONE_REDIRECT action is received" clause.

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
jrmykolyn: This update is complete, resolving.

packages/@storefront/flux-capacitor/test/unit/core/store/middleware.ts
jrmykolyn: This update is complete, resolving.

PR #101, opened 2019-05-06
SF-1359: Upgrade devDepednencies

jmbtutor:
This set will not result in a release.


jrmykolyn:
@jmbtutor 

Reviewed release notes for `nodemon`, `source-map-support`, and `webpack-cli`: no red flags. In terms of the `@types` packages, I agree with your statement that "if the project builds, they're safe." Approved.

PR #102, opened 2019-05-07
SF-1359: Upgrade TypeScript and typedoc

jmbtutor:
This will not trigger a release.


packages/@storefront/flux-capacitor/src/core/adapters/past-purchases.ts
```diff
@@ -30,7 +30,8 @@ namespace PastPurchasesAdapter {
     }, {});
   };
 
-  export const sortSkus = (skus: Store.PastPurchases.PastPurchaseProduct[], field: string) => {
+  // tslint:disable-next-line max-line-length
+  export const sortSkus = (skus: Store.PastPurchases.PastPurchaseProduct[], field: 'quantity'|'lastPurchased') => {
```
jrmykolyn: Do we use these values elsewhere? If so, should we make this a union type definition or enum?

package.json
```diff
@@ -56,10 +57,9 @@
     "ts-node": "^7.0.1",
     "tslint": "^5.11.0",
     "tslint-config-groupby": "^1.0.0",
-    "typedoc": "^0.13.0",
-    "typedoc-plugin-monorepo": "github:groupby/typedoc-plugin-monorepo",
+    "typedoc": "^0.14.2",
     "typedoc-plugin-nojekyll": "^1.0.1",
-    "typescript": "^3.1.3",
+    "typescript": "^3.4.5",
```
jrmykolyn: To confirm: Updating a dependency busts CircleCI's node dependency cache (ie. the `build_*` tests are passing using _this_ version of TypeScript)?

jmbtutor: Yes, CircleCI's cache is based on the hash of the `yarn.lock` file. The dependency changes, the hash changes, and the cache for the new hash doesn't exist. Plus, we're only caching yarn's global cache (`~/.cache/yarn`) and not the `node_modules`, so even if it were cached, the new version would get pulled in anyway.

jrmykolyn: Noted, thanks for clarifying. Marking this as resolved.

packages/@storefront/flux-capacitor/src/core/adapters/past-purchases.ts
jmbtutor: Just in this file and the related tests. I was hoping that there was a way to use the definition of `PastPurchaseProduct`, but I couldn't find one. I don't think there are enough uses to make a separate structure for this.

```
$ git grep -e "'lastPurchased'" -e "'quantity'"
flux-capacitor/src/core/adapters/past-purchases.ts:  export const sortSkus = (skus: Store.PastPurchases.PastPurchaseProduct[], field: 'quantity'|'lastPurchased') => {
flux-capacitor/src/core/adapters/past-purchases.ts:      case PAST_PURCHASE_SORTS.MOST_PURCHASED: return PastPurchasesAdapter.sortSkus(skus, 'quantity');
flux-capacitor/src/core/adapters/past-purchases.ts:      case PAST_PURCHASE_SORTS.MOST_RECENT: return PastPurchasesAdapter.sortSkus(skus, 'lastPurchased');
flux-capacitor/test/unit/core/adapters/past-purchases.ts:        expect(PastPurchaseAdapter.sortSkus(oldArr, 'quantity')).to.eql(newArr);
flux-capacitor/test/unit/core/adapters/past-purchases.ts:        expect(PastPurchaseAdapter.sortSkus(oldArr, 'lastPurchased')).to.eql([product3, product1, product2]);
flux-capacitor/test/unit/core/adapters/past-purchases.ts:      expect(sortSkusStub).to.be.calledWithExactly(skus, 'quantity');
flux-capacitor/test/unit/core/adapters/past-purchases.ts:      expect(sortSkusStub).to.be.calledWithExactly(skus, 'lastPurchased');
flux-capacitor/test/unit/core/reducers/data/past-purchases.ts:      expect(sorter).to.be.calledWithExactly(state.skus, 'quantity');
flux-capacitor/test/unit/core/reducers/data/past-purchases.ts:      expect(sorter).to.be.calledWithExactly(state.skus, 'lastPurchased');
```

jrmykolyn: @jmbtutor Since the goal of this ticket was to upgrade dependencies, I think we can leave the values as-is.

As an aside, do we degrade the development experience for consumers by *not* exposing these values as an enum of type definition?

jmbtutor: No. These particular values are only used in this file, and this module is an adapter for internal use. The only place this module gets used is in the reducer, and the values used to interact with this module are behind another enum-like type.

PR #103, opened 2019-05-08
SF-1153: Update tslint-config-groupby package version

JoeyDeol:
Bump tslint-config-groupby version to v1.1.0 and update tsconfig.json to accommodate tslint updates.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -6461,9 +6461,9 @@ tslib@^1.8.0, tslib@^1.8.1:
   integrity sha512-4krF8scpejhaOgqzBEcGM7yDIEfi0/8+8zDRZhNZZ2kjmHJ4hv3zCbQWxoJGz1iw5U0Jl0nma13xzHXcncMavQ==
 
 tslint-config-groupby@^1.0.0:
-  version "1.0.0"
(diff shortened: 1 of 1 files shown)


tsconfig.json
```diff
@@ -15,7 +15,9 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "outDir": "dist",
-    "sourceMap": true
+    "sourceMap": true,
+    "allowUnusedLabels": false,
+    "noImplicitReturns": true
```
jmbtutor: Remove this. I don't think we had the equivalent of this in the tslint config.

tsconfig.json
```diff
@@ -15,7 +15,9 @@
     "emitDecoratorMetadata": true,
     "experimentalDecorators": true,
     "outDir": "dist",
-    "sourceMap": true
+    "sourceMap": true,
+    "allowUnusedLabels": false,
```
jmbtutor: Isn't this already the default? https://www.typescriptlang.org/docs/handbook/compiler-options.html

tsconfig.json
JoeyDeol: Will do.

tsconfig.json
JoeyDeol: After review, we have decided to remove this all well.

